PR #13, opened 2026-08-09
Move plotting to Makie and fold the examples into the docs

michakraus:
## Summary

Plotting moves from Plots.jl to Makie and out of the package proper into a package extension, and the `examples/` notebooks are folded into the documentation.

Breaking: `plot(equ)` with Plots.jl no longer works. Version bumped to 0.8.0.

## Plotting

- The twelve `RecipesBase.@recipe` definitions scattered through `src/analytic/` are replaced by `ext/ElectromagneticFieldsMakieExt.jl`, loaded as soon as Makie or one of its backends is. `RecipesBase` and `LaTeXStrings` are no longer dependencies.
- `plot_equilibrium(equ; …)` creates a `Figure` and returns it; `plot_equilibrium!(fig[1,2], equ; …)` draws into an existing one at any Makie grid position and returns the `Axis`, or the vector of axes for the fields that draw more than one panel. The latter replaces Plots' `layout` for composing several equilibria into one figure.
- `plot_equilibrium` also takes `figure`, forwarded to `Makie.Figure`. Both it and `size` can carry a figure size; the precedence is the size chosen for the equilibrium, then `figure`, then an explicit `size`.
- Keyword arguments carry over (`nx`, `ny`, `levels`, `xlims`, `ylims`, `nτ`, `ni`), except that ABC's `nl` is now spelled `levels` like everywhere else, and `aspect` replaces `aspect_ratio`. Panels take an opt-in `colorbar`, the Solov'ev equilibrium a `boundary`; anything unrecognised is forwarded to `contour!`.
- Equilibria without a plotting method — the three Penning traps — report that in an `ArgumentError` rather than a `MethodError` on an internal helper.

## Bugs found during the port

- **Six of the contour plots were transposed.** Plots.jl expects the value matrix indexed as `z[j,i]` for `(x[i], y[j])`, and only the tokamak and Solov'ev recipes transposed accordingly — ABC, Dipole, QuadraticPotentials, Singular, SymmetricQuadratic and ThetaPinch plotted the mirror image about the diagonal. Those six are exactly the six with square default grids, `nx == ny == 100`, which is why Plots.jl never raised a dimension error. Makie uses the `z[i,j]` convention the comprehensions already produce, and the suite now plots every field on a non-square grid so a reintroduced transpose fails loudly.
- **The tokamak and Solov'ev plots showed the wrong quantity.** They contoured `A₃ / R`, the *physical* toroidal component, and described it as the poloidal flux function. The flux function is the covariant component `A₃ = ψ`: `B · ∇A₃` vanishes while `B · ∇(A₃/R)` does not, so the contours drawn were not flux surfaces. Only the `ψ = 0` contour was unaffected, which is why the red plasma boundary always looked right. The cartesian tokamak now plots `R A_y`, its `A_y` at `y = 0` being that same physical component, so all three tokamak charts draw the same picture. `test_analytic.jl` asserts the invariant, including the stand-in that fails it.
- **`SolovevSymmetric`'s default plot range was centred on the wrong side.** Its flux function depends on `x` through `(R₀ + x)⁴`, so the magnetic axis sits at `x = -R₀` while the window was centred on `+R₀`; for any `R₀ ≠ 0` the plot showed a monotone ramp rather than closed flux surfaces.
- The quantity plotted as `|B|` for the ABC field was `|B|²`, and the one plotted as `B_z` for the symmetric quadratic field was `B₀/(1+x²+y²)` where the field is `B₀(1+x²+y²)`. Both helpers are used only for plotting; the generated evaluation routines were never affected.
- The logarithmic levels of the singular field are now taken over the finite magnitudes, which copes both with a grid that meets the singular line and with a window in which a component keeps one sign.
- **`Documenter` was a dependency of the package proper.** `src/ElectromagneticFields.jl` carried a `using Documenter` that nothing used, so every downstream install pulled in its tree.
- **The matrix wrappers allocated about seven times the matrix they return.** A generated body that does not mention the coordinates evaluated to whatever literal type SymEngine emitted, so the structurally constant components came out as `Int64` — `g₁₁`, the off-diagonal entries of `g` and `DF`, `φ` and `E` of a purely magnetic equilibrium — while the rest were `Float64`, and the mixture made `hvcat` promote at runtime. `g` allocated 976 bytes and `DF` 1072, where the 3×3 matrix is 144. Both are now 144, the scalar functions still allocate nothing, and `test_analytic.jl` asserts as much for every equilibrium. The one behavioural difference is that a coordinate-independent constant such as `B₃ = B₀R₀` now takes the coordinate's type.

## Documentation

- `docs/src/analytic.md` is split into one page per field under `docs/src/analytic/`.
- The hidden `@eval` + `savefig` + image-link pattern gives way to `@example`, so the code that produces each figure is visible and is executed at build time. The figures are rendered as SVG, which suits line contours and silences a Documenter warning that covered every figure block.
- New `usage.md` covers what the docs never explained: `@code` vs `code` vs `load_equilibrium`, why the do-block form exists, the generated API, and grid evaluation. New `plotting.md` documents the plotting interface.
- **Coordinates** — what a chart is and how one is declared, the tangent map and the metric, the three component representations the package carries with the conversions between them, and the volume element together with the orientation that distinguishes it from the signed determinant. It states plainly that "physical" here means the components in the ambient cartesian frame, not the `√gᵢᵢ vⁱ` normalisation that some texts use.
- **Fields** — the vector potential as the only magnetic input, the magnetic field in its four guises, the orthonormal frame along it, the electric field, and the fact that none of it depends on the chart. Two naming traps are called out: `B` is the scalar `|B|`, and `B²` is the second contravariant component rather than the two-form of the same name in the formulas. The Hodge star formula uses the signed determinant `det DF` rather than `√g`, since the star is orientation-dependent.
- `docs/src/modules.md` filters `plot_equilibrium` and `plot_equilibrium!` out of its `@autodocs` block, since the Plotting page documents them in context. `docs/Project.toml` swaps Plots for CairoMakie, pins both it and Documenter, and resolves the package through a `[sources]` entry rather than a `Pkg.develop` in CI.
- Every identity asserted in the prose is checked by an `@example` block carrying a hidden `@assert`, so an identity that stops holding fails the build instead of quietly rendering `false`.

### Corrections to the new prose and to older docstrings

- The `φ` in the three Penning trap docstrings was missing its minus sign; the potential is `-E₀(x²/2 + y²/2 - z²)`, which is what makes the stated `E = E₀(x, y, -2z)` follow and what the code computes.
- The ABC page had the integrability criterion backwards: the field lines are integrable when one of `a`, `b`, `c` vanishes, and `a = b = c = 1` — the package default — is the classic chaotic case.
- The singular field's vector potential diverges as `r⁻²`, not `r⁻³`; only `|B|` goes as `r⁻³`.
- Four of the five charts are left-handed, not three.
- `plotting.md` promised `size` to `plot_equilibrium!`, which errors on it, and plotting methods to every equilibrium.
- `SolovevSymmetric` is documented as cartesian, and its `R₀` as the parameter that puts the magnetic axis at `x = -R₀`.

### Contour levels

With `ψ` no longer divided by `R`, its range over the plot window is dominated by the far field, where it grows without bound. The Solov'ev levels are therefore anchored to the flux on the magnetic axis rather than spread over the sampled range: uniform in `ψ`, one level exactly on the plasma boundary, and a quarter of them inside it — ten at the default `levels = 40`. Passing the levels themselves still bypasses this.

## Infrastructure

- The `Documentation` workflow drops four redundant steps — a `Pkg.develop` duplicating the `[sources]` entry in `docs/Project.toml`, a `Pkg.build`/`Pkg.precompile` pair covered by `Pkg.instantiate`, a `julia-buildpkg` for the main project the doc build never uses, and a standalone doctest run repeating what `makedocs` already does — pins Julia to 1.12 as `[sources]` requires, and caches its depot.
- The stale Travis and Coveralls badges are gone; the CI and Codecov badges in the documentation now match the working ones in the README, with a Documentation badge added to both.

## Review status

Every item from the two reviews on d868543 is addressed and its thread resolved — the two Copilot
comments and items 1 to 6, plus the minor points and the out-of-scope `Documenter` note, across
e61633e, debebdb and f00922d. All 13 CI jobs pass on the head commit, and the `Documentation`
workflow builds on Julia 1.12 and correctly declines to deploy from a pull request.

One thing is deliberately left, and it does not block: the three-panel Solov'ev comparison figure is
still duplicated verbatim in `docs/src/plotting.md` and `docs/src/analytic/solovev.md`. It earns its
place on both pages — one as the worked example of composing a figure, the other as the tour of the
named configurations — and Documenter has no include mechanism that would share the block without
hiding the code, which is the point of using `@example` here.

Two further nits raised in review are fixed in f00922d: `plot_equilibrium` gives `size` and `figure`
a defined precedence rather than letting a size inside `figure` win by accident, documented and
pinned by a test; and the commentary on common subexpression elimination in
`src/analytic/analytic_equilibrium.jl` describes the alternatives as alternatives instead of as
before-and-after notes.

## Removed

`docs/travis.sh`, a deploy script for a CI service the repository stopped using years ago, along with the Travis badge it belonged to.

`examples/` in its entirety. The notebooks predated the current API, carried no narrative, and `axisymmetric_tokamak.ipynb` had been sitting in the repository with unresolved merge-conflict markers (invalid JSON) since 2020. What was worth keeping is in the per-field pages, ported to the current API.

## Test plan

- [x] `Pkg.test()` passes, including a new `test/test_plots.jl` covering every plotting method, keyword overrides, figure composition, all three grid-position types, and that the Penning traps correctly have none — nothing tested plotting before.
- [x] `test_analytic.jl` gains three guards: `A₃` is a flux label for the axisymmetric equilibria (and `A₃/R` is not), every component of an equilibrium returns one type, and the scalar functions allocate nothing while the matrix wrappers allocate only their matrix.
- [x] `julia --project=docs docs/make.jl` builds clean: 23 figures, no Documenter warning other than the deployment auto-detect. Verified separately that a deliberately failing `@assert` in an `@example` block does terminate the build.
- [x] A clean `Pkg.instantiate()` on the docs environment resolves the package from `[sources]` with no `Pkg.develop`, and `Documenter` is absent from the package's dependency tree.
- [x] Each figure compared against the corresponding pre-existing one; the transposition and flux-function fixes are the only intended differences, and the Solov'ev plasma boundary still lies on a contour.


ext/ElectromagneticFieldsMakieExt.jl
```diff
@@ -0,0 +1,359 @@
...
+    ygrid = grid(ylims, ny)
+
+    pot1 = [A₁([xgrid[i], ygrid[j], 0.0], equ) for i in eachindex(xgrid), j in eachindex(ygrid)]
+    pot2 = [A₂([xgrid[i], ygrid[j], 0.0], equ) for i in eachindex(xgrid), j in eachindex(ygrid)]
+    Bfield = [ElectromagneticFields.Singular.B([xgrid[i], ygrid[j], 0.0], equ) for i in eachindex(xgrid), j in eachindex(ygrid)]
+
+    for (n, (vals, title, lvls)) in enumerate((
+        (pot1, L"A_x (x,y)", doublelogrange(0.1, maximum(pot1), levels)),
+        (pot2, L"A_y (x,y)", doublelogrange(0.1, maximum(pot2), levels)),
+        (Bfield, L"B_z (x,y)", logrange(max(0.1, minimum(Bfield)), maximum(Bfield), levels)),
```
michakraus: **1. Confirmed, and there are two failure modes here, not one.** Copilot flagged the non-finite case; the sign case is the one that bites more easily.

```julia
julia> plot_equilibrium(Singular.init(); nx=101, ny=101)
ERROR: ArgumentError: start and stop must be finite, got -1.0 and NaN
```
An odd grid over `(-1,1)` lands exactly on the origin, where `A₁ = 0/0 = NaN` and `B = Inf`. The defaults `nx = ny = 100` miss it, so this is invisible until someone picks an odd resolution.

```julia
julia> plot_equilibrium(Singular.init(); xlims=(2.0,3.0), ylims=(2.0,3.0))
ERROR: DomainError with -0.03928371006591932:
log10 was called with a negative real argument ...
```
This one needs no singularity at all: `A₂ = -B₀x/r³` is negative everywhere for `x > 0`, so `maximum(pot2) < 0` and `log10` throws. Any window that does not straddle the axis fails.

Deriving the extent from the finite magnitudes fixes both:

```julia
function logextent(vals; lo=0.1)
    hi = maximum(abs, filter(isfinite, vals); init=lo)
    max(hi, 10lo)          # keep the log range non-degenerate
end
```

then `doublelogrange(0.1, logextent(pot1), levels)`, and the same clamp for `Bfield`.

While here: `doublelogrange` returns `[-0.1, …, -max, +0.1, …, +max]`, which is not monotonic. Makie stores the vector verbatim (I checked `plot.levels[]`), so the lines still draw, but the colormap is sampled against an unsorted range. A `sort!` costs nothing. Same in the old Plots recipe, so not a regression — just worth fixing while the code is being moved.

ext/ElectromagneticFieldsMakieExt.jl
```diff
@@ -0,0 +1,359 @@
...
+
+figuresize(::ABCEquilibrium) = (400, 1200)
+
+function plot_equilibrium!(position::Position, equ::ABCEquilibrium;
+    nx=99, ni=div(nx, 2) + 1, levels=12, kwargs...)
+
+    lims = (0, 2π)
+    xgrid = grid(lims, nx)
+    Bfield = [ElectromagneticFields.ABC.B([xgrid[i], xgrid[j], xgrid[k]], equ)
+              for i in eachindex(xgrid), j in eachindex(xgrid), k in eachindex(xgrid)]
```
michakraus: **2. Confirmed — real, though modest at the default.** Measured on this branch at `nx = 99`: 0.22 s and 7.8 MB to build the cube, of which `3nx²` of `nx³` values (about 3%) are ever read. It is the `n³` scaling that makes it worth fixing rather than the default cost — roughly 1.8 s / 62 MB at `nx = 200`, 15 s / 0.5 GB at `nx = 400`.

Three 2-D comprehensions at `xgrid[ni]` give the same three panels:

```julia
Bxy = [B([xgrid[i], xgrid[j], xgrid[ni]], equ) for i in eachindex(xgrid), j in eachindex(xgrid)]
Bxz = [B([xgrid[i], xgrid[ni], xgrid[k]], equ) for i in eachindex(xgrid), k in eachindex(xgrid)]
Byz = [B([xgrid[ni], xgrid[j], xgrid[k]], equ) for j in eachindex(xgrid), k in eachindex(xgrid)]
```

Separately, the titles hardcode `\pi`, but `ni = div(nx,2)+1` only lands on π for odd `nx`: with `nx = 100`, `xgrid[51] ≈ 3.17` and the panel is labelled `|B(x,y,π)|` regardless. Either interpolate `xgrid[ni]` into the title or say in the docstring that `nx` should be odd. (Inherited from the old recipe.)

ext/ElectromagneticFieldsMakieExt.jl
```diff
@@ -0,0 +1,359 @@
...
+    title="", xlabel=L"x", ylabel=L"y", levels=20,
+    aspect=DataAspect(), colorbar=false, colormap=:viridis, kwargs...)
+
+    ax = Axis(position[1, 1]; title=title, xlabel=xlabel, ylabel=ylabel, aspect=aspect)
+    contour!(ax, xgrid, ygrid, vals; levels=levels, colormap=colormap, kwargs...)
+
+    if colorbar
+        # a line contour carries no single colormap that Makie could derive a
+        # colorbar from, so it is built from the range of the data instead
+        Colorbar(position[1, 2]; colormap=colormap, limits=extrema(filter(isfinite, vals)))
```
michakraus: **3.** `extrema(filter(isfinite, vals))` throws `ArgumentError: reducing over an empty collection is not allowed` when nothing in the panel is finite. Reachable via the Singular field on a grid that hits the axis, combined with `colorbar = true`. An `init` on the reduction, or skipping the colorbar when there is no finite data, is enough.

ext/ElectromagneticFieldsMakieExt.jl
```diff
@@ -0,0 +1,359 @@
...
+    for (n, (vals, xlabel, ylabel, title)) in enumerate((
+        (Bfield[:, :, ni], L"x", L"y", L"|B(x,y,\pi)|"),
+        (Bfield[:, ni, :], L"x", L"z", L"|B(x,\pi,z)|"),
+        (Bfield[ni, :, :], L"y", L"z", L"|B(\pi,y,z)|"),
+    ))
+        contourpanel!(position[n, 1], xgrid, xgrid, vals;
+            panelopts(kwargs; title=title, xlabel=xlabel, ylabel=ylabel, levels=levels)...)
+    end
+
+    position
```
michakraus: **4. API question, and the one thing I would settle before tagging 0.8.0** — because a return value is breaking to change afterwards.

Every `plot_equilibrium!` method returns `position`, which the caller already holds. Returning the `Axis` (or the vector of axes for the multi-panel fields) would let callers do the things one actually wants after drawing: `xlims!`, overlay a particle trajectory, attach a legend. `contourpanel!` already returns `ax`, so it is a mechanical change.

Related, and cheap: `plot_equilibrium(PenningTrapUniform.init())` currently reports

```
MethodError: no method matching figuresize(::PenningTrapUniformEquilibrium{Float64})
```

`figuresize` is an internal helper the user has never heard of, and the failure is really "this field has no plotting method". A fallback

```julia
figuresize(equ::ElectromagneticFields.AnalyticEquilibrium) =
    error("no plotting method is defined for $(typeof(equ)); see the Plotting page for how to plot it by hand")
```

says so directly. `test/test_plots.jl` would need its `@test_throws MethodError` adjusted.

src/plots.jl
```diff
@@ -0,0 +1,56 @@
...
+plot_equilibrium(Solovev.ITER())
+```
+
+Which quantity is shown depends on the equilibrium: for the tokamak and Solov'ev
+equilibria it is the poloidal flux function ``A_\phi / R``, for the fields defined
+in cartesian coordinates it is a panel per vector potential component, and for the
+ABC field it is the absolute value of the magnetic field in three mid-planes.
+
+All methods accept the resolution of the evaluation grid (`nx`, `ny`), the number of
+contour `levels`, the plot ranges `xlims` and `ylims`, and the figure `size`.
```
michakraus: **5.** "All methods accept … `nx`, `ny` … `xlims` and `ylims` … and the figure `size`" is not quite true: ABC accepts none of `ny`, `xlims`, `ylims` (its domain is fixed to `[0,2π]²`), and `plot_equilibrium!` accepts no `size` at all. `docs/src/plotting.md` line 23 makes the same claim.

Also undocumented here: `colorbar`, `aspect`, `ni` (ABC), and `nτ` / `boundary` (Solov'ev) — `boundary` in particular is a new, useful knob that appears nowhere in the docs or the changelog. Either enumerate the per-field keywords or soften to "most methods accept".

test/test_plots.jl
```diff
@@ -0,0 +1,49 @@
...
+        @test plot_equilibrium(equ) isa Makie.Figure
+    end
+
+    # keywords are forwarded to the axis and the contour call, and the per-panel
+    # defaults must not collide with what the caller passes
+    for equ in plot_equilibria
+        @test plot_equilibrium(equ; levels=5, colorbar=true, colormap=:plasma) isa Makie.Figure
+    end
+
+    # drawing into an existing figure, which is how several equilibria are composed
```
michakraus: **6.** Good to finally have coverage here. Two gaps, the first of which matters because it is the bug this PR fixes:

**The suite cannot catch a re-introduced transpose.** Every cartesian field defaults to `nx == ny == 100`, so a transposed value matrix is still square and Makie accepts it — which is exactly how the bug survived under Plots.jl. Adding a case with `nx != ny` closes that off, because Makie rejects a matrix whose dimensions do not match `(length(x), length(y))` (I confirmed it errors):

```julia
for equ in plot_equilibria
    @test plot_equilibrium(equ; nx=37, ny=53) isa Makie.Figure
end
```
(ABC would need excluding, since it takes no `ny`.)

**Also worth adding:**
- the two Singular cases from the comment above (`nx=101`, and `xlims` off the axis);
- `plot_equilibrium!` into a `GridLayout` and a `GridSubposition` — both are in the `Position` union but only `GridPosition` is exercised;
- an `xlims`/`ylims` override, which no test passes at all.

michakraus (commented):
## Review

I checked the claims in the description against the branch and ran the plotting code; the two Copilot comments are both real, and the first one has a second failure mode Copilot did not mention. Details inline. Nothing here blocks the direction of the PR — the extension is a clear improvement over twelve scattered recipes — but items 1, 2 and 4 are worth doing before tagging 0.8.0.

### What checks out

- **The transposition fix is real, and the evidence is stronger than the description suggests.** The six recipes that did not transpose (ABC, Dipole, QuadraticPotentials, Singular, SymmetricQuadratic, ThetaPinch) are *exactly* the six with square default grids, `nx == ny == 100`. That is why Plots.jl never raised a dimension error and the bug survived for years — the two families that used `nx=100, ny=120` are precisely the two that got `pot'` right. Worth adding to the changelog entry; it also dictates how to test for a regression (see item 6).
- **Both field-helper fixes are correct.** For `SymmetricQuadratic`, `B_z = ∂ₓA₂ − ∂_yA₁ = B₀(2+3x²+y²)/4 + B₀(2+x²+3y²)/4 = B₀(1+x²+y²)`, matching the module docstring. For ABC, `|B| = √(A₁²+A₂²+A₃²)` since `B = A`. And neither `B` appears in the corresponding `get_functions`, so the claim that the generated evaluation routines were never affected holds.
- No stale references to `Plots`, `RecipesBase`, `LaTeXStrings`, `examples/` or `analytic.md` anywhere in `src/`, `docs/`, `test/`, `.github/` or the README. `Manifest.toml` is gitignored and untracked.
- The physics/orientation claims in the new pages spot-check out — e.g. `AxisymmetricTokamakToroidal`'s `orientation(), det(DF(t,ξ)) ≈ orientation()*J(t,ξ)` evaluates to `(-1, true)`.

### Minor points not worth an inline comment

- `docs/Project.toml` gained a `[sources]` section, which Pkg only understands from Julia 1.11, while the package declares `julia = "1.10"`. `.github/workflows/Documentation.yml` still does `Pkg.develop(PackageSpec(path=pwd()))`, which is now redundant with it. Harmless on CI (`setup-julia` takes the latest release) — but pick one of the two.
- `plot_equilibrium(equ; figure = (; size = (300,300)))` silently overrides the `size` argument, because `Figure(; size = size, figure...)` lets the splat win. Either document that or fold `size` into `figure`.
- The three-panel Solov'ev comparison figure is duplicated verbatim in `docs/src/plotting.md` and `docs/src/analytic/solovev.md`.
- Dropping `using RecipesBase` left a stray double blank line after `module ABC`, `module SymmetricQuadratic`, `module Singular` and friends.

### Out of scope, but noticed

`src/ElectromagneticFields.jl:3` has `using Documenter`, so `Documenter` sits in `[deps]` and every downstream user installs it. Pre-existing; the `[targets] docs = [...]` entry is likewise inert, since Pkg only honours the `test` target. Separate issue rather than growing this PR.

**Update:** folded into this PR after all, in debebdb. `using Documenter` is gone from `src/ElectromagneticFields.jl`, along with the `[deps]` and `[compat]` entries and the inert `[targets] docs` line, so Documenter no longer appears in a downstream install's dependency tree. `docs/Project.toml` declares it for the doc build, which is the only thing that needs it.

ext/ElectromagneticFieldsMakieExt.jl
```diff
@@ -0,0 +1,359 @@
...
+              for i in eachindex(xgrid), j in eachindex(xgrid), k in eachindex(xgrid)]
+
+    for (n, (vals, xlabel, ylabel, title)) in enumerate((
+        (Bfield[:, :, ni], L"x", L"y", L"|B(x,y,\pi)|"),
+        (Bfield[:, ni, :], L"x", L"z", L"|B(x,\pi,z)|"),
+        (Bfield[ni, :, :], L"y", L"z", L"|B(\pi,y,z)|"),
+    ))
+        contourpanel!(position[n, 1], xgrid, xgrid, vals;
+            panelopts(kwargs; title=title, xlabel=xlabel, ylabel=ylabel, levels=levels)...)
+    end
```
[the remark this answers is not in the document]
michakraus: Confirmed — measured at the default `nx = 99`: 0.22 s and 7.8 MB for the cube, of which about 3% (`3nx²` of `nx³`) is read. The `n³` scaling is the real problem: ~1.8 s / 62 MB at `nx = 200`, ~15 s / 0.5 GB at `nx = 400`. Will replace with three 2-D comprehensions at `xgrid[ni]`. See the review comment above for details.

ext/ElectromagneticFieldsMakieExt.jl
```diff
@@ -0,0 +1,359 @@
...
+
+    pot1 = [A₁([xgrid[i], ygrid[j], 0.0], equ) for i in eachindex(xgrid), j in eachindex(ygrid)]
+    pot2 = [A₂([xgrid[i], ygrid[j], 0.0], equ) for i in eachindex(xgrid), j in eachindex(ygrid)]
+    Bfield = [ElectromagneticFields.Singular.B([xgrid[i], ygrid[j], 0.0], equ) for i in eachindex(xgrid), j in eachindex(ygrid)]
+
+    for (n, (vals, title, lvls)) in enumerate((
+        (pot1, L"A_x (x,y)", doublelogrange(0.1, maximum(pot1), levels)),
+        (pot2, L"A_y (x,y)", doublelogrange(0.1, maximum(pot2), levels)),
+        (Bfield, L"B_z (x,y)", logrange(max(0.1, minimum(Bfield)), maximum(Bfield), levels)),
+    ))
```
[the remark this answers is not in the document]
michakraus: Confirmed, and there is a second failure mode you did not mention that needs no singular line at all: `plot_equilibrium(Singular.init(); xlims=(2.0,3.0), ylims=(2.0,3.0))` throws a `DomainError` in `log10`, because `A₂ = -B₀x/r³` is negative everywhere for `x > 0` and so `maximum(pot2) < 0`. The non-finite case reproduces as you describe — `nx = ny = 101` hits the origin and gives `ArgumentError: start and stop must be finite, got -1.0 and NaN`. Fixing both by taking the extent from `maximum(abs, filter(isfinite, vals))`. See the review comment above.

ext/ElectromagneticFieldsMakieExt.jl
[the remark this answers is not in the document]
michakraus: Fixed in e61633e. The cube is gone; the three panels come from three 2-D comprehensions evaluated at `xgrid[ni]`, so `3nx²` values are computed instead of `nx³`.

ext/ElectromagneticFieldsMakieExt.jl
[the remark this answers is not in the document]
michakraus: Fixed in e61633e. The extent of the level range now comes from `logextent`, which reduces `maximum(abs, …)` over `Iterators.filter(isfinite, vals)` with an `init`, so neither `Inf`/`NaN` on the singular line nor a component of one sign can reach `log10`. Both cases are in `test/test_plots.jl`.

ext/ElectromagneticFieldsMakieExt.jl
michakraus: Fixed in e61633e, both failure modes. `logextent(vals; lo)` takes the upper end from `maximum(abs, Iterators.filter(isfinite, vals); init = lo)` and clamps it to `max(hi, 10lo)`, which covers the odd-grid `NaN` case and the negative-maximum case alike. `doublelogrange` now ends in `sort!`, so the vector Makie stores is monotonic. `test/test_plots.jl` exercises `nx = ny = 101` and `xlims = ylims = (2.0, 3.0)`.

ext/ElectromagneticFieldsMakieExt.jl
michakraus: Fixed in e61633e. Three 2-D comprehensions at `xgrid[ni]` replace the cube.

On the titles: taken the docstring route rather than interpolating. `plot_equilibrium!` documents `ni` as "the index of the mid-plane, by default the grid point closest to ``\pi`` (exactly ``\pi`` for odd `nx`)", and the default `nx = 99` is odd, so the labels are exact unless a caller asks for an even grid.

ext/ElectromagneticFieldsMakieExt.jl
michakraus: Fixed in e61633e. `contourpanel!` filters first and skips the colorbar entirely when nothing is finite:

```julia
finite = filter(isfinite, vec(vals))
if !isempty(finite)
    lo, hi = extrema(finite)
    lo == hi && ((lo, hi) = (lo - one(lo), hi + one(hi)))
    Colorbar(position[1, 2]; colormap = colormap, limits = (lo, hi))
end
```

The `lo == hi` line covers the panel that is constant across the grid, which Makie also rejects. `plot_equilibrium(Singular.init(); nx = 101, ny = 101, colorbar = true)` is in the test suite.

ext/ElectromagneticFieldsMakieExt.jl
michakraus: Both fixed in e61633e, before tagging.

`plot_equilibrium!` returns the `Axis`, or the `Vector{Axis}` for the fields that draw a panel per component; `plot_equilibrium` still returns the `Figure`. `test/test_plots.jl` asserts it for every equilibrium, and `docs/src/plotting.md` uses the returned axis to mark the magnetic axis with `scatter!`.

`figuresize` now carries a fallback that throws `ArgumentError("no plotting method is defined for $(typeof(equ)); see the Plotting page …")`, and the three Penning trap tests assert `ArgumentError` rather than `MethodError`.

src/plots.jl
michakraus: Fixed across e61633e and debebdb. `src/plots.jl` now splits the list: `levels`, `title`/`xlabel`/`ylabel`, `aspect` and `colorbar` are common to every method, `nx`/`ny`/`xlims`/`ylims` are qualified with "most methods", and the two exceptions are spelled out — ABC with its cubic grid and `ni`, Solov'ev with `boundary` and `nτ`. `size` is documented as belonging to `plot_equilibrium` alone.

`docs/src/plotting.md` still claimed `size` for every method after e61633e; debebdb fixes that too, adds `colorbar` to the common list, and drops the assertion that every equilibrium can be plotted. It also documents how an integer `levels` is interpreted for the Solov'ev equilibria, which changed in the same commit.

test/test_plots.jl
michakraus: Fixed in e61633e; all four gaps are covered.

- `plot_equilibrium(equ; nx = 37, ny = 53)` for every rectangular field, ABC excluded as you note. That is the one that matters — it is what would catch a reintroduced transpose.
- The two Singular cases, `nx = ny = 101` and `xlims = ylims = (2.0, 3.0)`, plus the first of them again with `colorbar = true` for the empty-reduction path.
- `plot_equilibrium!` into a `GridSubposition` (`fig[1,2][1,1]`) and a `GridLayout`, alongside the `GridPosition` that was already there.
- An `xlims`/`ylims` override for every rectangular field.

`debebdb` adds three more guards in `test_analytic.jl`, of which one is in the same spirit as the transpose test: it asserts that `A₃` is a flux label for the axisymmetric equilibria and that `A₃ / R` is not, which is what pins down the quantity the Solov'ev and tokamak panels contour.

PR #14, opened 2026-09-08
Normalise the analytic equilibrium sources to NFC

michakraus:
Converts `src/analytic/analytic_equilibrium.jl` and `test/test_analytic.jl` from NFD to Unicode NFC. They stored `ḡ` (21×), `â` (3×) and `ĉ` (3×) as a base letter plus a combining mark — an artefact of macOS, not a decision.

Part of a tree-wide conversion (`Tasks/Normalise the Julia sources to NFC.md`).

## Why, and the hazard this actually removes

Julia's parser normalises identifiers to NFC, so nothing about the compiled code changes. The cost of NFD is in tooling: a `grep` pattern or an editor search typed in NFC matches nothing in such a file, silently.

But this file had a sharper problem, worth naming because it is exactly the trap the conversion exists to close: **it already spelled the same glyph both ways.** `functions["ḡ"]` at `src/analytic/analytic_equilibrium.jl:824` was *precomposed* on `main`, while the `ḡ` identifiers around it were *decomposed*. That worked only because the parser normalises the identifier side and not the string side. Nothing was broken, but the two encodings were one careless edit apart from disagreeing.

After this branch, **zero non-NFC string literals remain under `src/`**.

`DF̄` is safe either way — a macron over `F` has no precomposed codepoint.

## The one runtime-visible change

`"Dḡ"` and `"DDḡ"` at `:286,292` are string literals, which are *not* parser-normalised, so their bytes change. They are labels passed to `symprint`, which only `println`s them (`:953`). Neither string occurs anywhere in `src/`, `test/` or `docs/`, so nothing reads that output back.

## Verification

| what | result |
|:--|:--|
| diff is exactly the NFC normalisation | 2/2 files |
| exhaustive `Expr`-leaf diff | **exactly two** differences, both the `symprint` labels |
| `test/test_analytic.jl` | `Expr`-identical |
| `Pkg.test()` | **30 testsets, 0 failed** (incl. 8824 CSE-preservation assertions) |
| no `functions[...]` Dict key is NFC-changing | proven by the exhaustiveness of the leaf diff |
| `fatou lint` | findings identical base vs head |

The leaf diff being *exhaustive* is what makes the Dict-key claim a proof rather than a sample: a changed key would have shown up as a third difference.

**Not applicable** rather than unmeasured: piracy, inference, allocations — apart from the two printed labels, base and head parse to identical trees.

## Pre-existing, deliberately not addressed

`src/analytic/analytic_equilibrium.jl` uses **two visually indistinguishable identifiers for the inverse metric**: `g̅` (`g` + combining overline, 15×, lines 141–206) and `ḡ` (`g` + macron, lines 285–513, 824). They are distinct bindings, always have been, and the tests are green — but nothing on screen tells a macron from an overline. Renaming one is a design decision and would falsify this branch's byte-equality claim, so it wants its own change.

Also: `unused-binding` at `:243` (`t` assigned, never used), 42 lines before the first hunk.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,18 @@ not covered here; see the git history for those.
 - `src/analytic/analytic_equilibrium.jl` and `test/test_analytic.jl` are now Unicode
   NFC-normalised. They stored `ḡ` (21 times), `â` (3) and `ĉ` (3) as a base letter plus a combining
   mark, inherited from macOS rather than chosen, which makes no difference to the compiled code —
-  Julia's parser normalises
-  identifiers to NFC — but defeats every byte-matching tool: a `grep` pattern or an editor search
-  typed in NFC matches nothing in such a file, silently.
+  Julia's parser normalises identifiers to NFC — but defeats every byte-matching tool: a `grep`
+  pattern or an editor search typed in NFC matches nothing in such a file, silently.
+
+  Worth naming, because it is the hazard this removes: the file already spelled the same glyph
+  both ways. `functions["ḡ"]` was precomposed while the `ḡ` identifiers around it were decomposed,
+  and that worked only because the parser normalises the identifier side and not the string. No
+  non-NFC string literal remains under `src/`. `DF̄` is safe either way — a macron over `F` has no
+  precomposed codepoint.
+
+  A separate note for a reader of this file: `g̅` elsewhere in the same source is a *different*
+  binding, `g` plus a combining overline rather than a macron. It is unchanged, and
+  indistinguishable from `ḡ` on screen.
 
   The only bytes that change at runtime are the `"Dḡ"` and `"DDḡ"` labels passed to `symprint`,
   which prints them; string literals are not parser-normalised. Nothing reads that output back.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html). Releases 
 not covered here; see the git history for those.
 
 
+## [Unreleased]
+
+### Changed
+
+- `src/analytic/analytic_equilibrium.jl` and `test/test_analytic.jl` are now Unicode
+  NFC-normalised. They stored `ḡ` as a base letter plus a combining mark, inherited from macOS
+  rather than chosen, which makes no difference to the compiled code — Julia's parser normalises
+  identifiers to NFC — but defeats every byte-matching tool: a `grep` pattern or an editor search
+  typed in NFC matches nothing in such a file, silently.
+
+  The only bytes that change at runtime are the `"Dḡ"` and `"DDḡ"` labels passed to `symprint`,
+  which prints them; string literals are not parser-normalised. Nothing reads that output back.
+
 ## [0.8.0] - 2026-08-10
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,9 @@ not covered here; see the git history for those.
 ### Changed
 
 - `src/analytic/analytic_equilibrium.jl` and `test/test_analytic.jl` are now Unicode
-  NFC-normalised. They stored `ḡ` as a base letter plus a combining mark, inherited from macOS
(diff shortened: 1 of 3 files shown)


michakraus (commented):
**Verdict: approve** — this is exactly what it claims to be, and every claim in the PR body was independently reproduced.

## Scope

`CHANGELOG.md`, `src/analytic/analytic_equilibrium.jl`, `test/test_analytic.jl` — confirmed via `git diff --name-status main...origin/pr/14`, matching the PR body's file list.

## Verification performed

**Glyph set and counts.** Scanned both changed `.jl` files for base+combining-mark pairs that recompose under NFC (a Julia script, not grep, so it can't miss a decomposed occurrence). Result: `ḡ` × 8 in `analytic_equilibrium.jl` + `ḡ` × 13, `â` × 3, `ĉ` × 3 in `test_analytic.jl` — totals `ḡ` 21, `â` 3, `ĉ` 3, exactly matching the CHANGELOG entry and PR body. Head-file scan found zero remaining decomposed pairs in either file.

**Parse-tree equality.** `Meta.parseall` on base vs. head:
- `test/test_analytic.jl`: trees `==` (fully identical, 5540/5540 leaves).
- `src/analytic/analytic_equilibrium.jl`: trees differ at exactly 2 of 3517 leaves — `"Dḡ"` and `"DDḡ"`, the two `symprint` string-literal labels. Matches the PR's claimed "exactly two differences."

**The one runtime-visible change.** `symprint(name, symexpr, output=1, detail_level=0)` (`src/analytic/analytic_equilibrium.jl:953-957`) only does `println(name, " = ", symexpr, "\n")` — no return value, nothing downstream consumes it. Grepped `src/` and `test/` for both the NFC and NFD spellings of `"Dḡ"`/`"DDḡ"`: no other occurrence exists anywhere in the tree. Confirmed clean.

**The `functions["ḡ"]` mixed-encoding claim** (line 824): read the raw codepoints — `U+1E21` (LATIN SMALL LETTER G WITH MACRON, precomposed), confirming this literal was already NFC on `main` while the surrounding `ḡ` identifiers were decomposed, exactly as described. It is untouched by this diff (not in either hunk), consistent with needing no change.

**Doctests / fenced blocks.** No `jldoctest` blocks anywhere in `docs/`, no `.jmd` files in the repo, and nothing in `docs/` references `generate_equilibrium_functions` or `analytic_equilibrium` at all — so there is no doc content that could be affected regardless.

**`fatou lint`.** Ran on both base and head copies of the two changed files: identical single finding both times — `unused-binding` at `analytic_equilibrium.jl:243` (`t` assigned, never used across the whole enclosing function, verified by scanning through its `end` at line 521). Pre-existing, 42 lines before the first hunk, honestly disclosed in the PR body as deliberately not addressed. No new finding introduced.

**CHANGELOG structure.** `## [Unreleased]` / `### Changed` matches the shape of the following `## [0.8.0]` section exactly; single `[Unreleased]` heading (no duplicate); blank line separates it from `## [0.8.0]` below. No "no string literal was affected" claim is made — the entry correctly states the opposite (the two `symprint` labels do change bytes) and names why nothing reads them back.

**NFC invariant.** `s == Unicode.normalize(s, :NFC)` holds for all three changed files at head.

**CI.** `gh pr checks 14`: all 10 jobs pass, including a dedicated "Doctests" job (56s) and "Documentation" (11m24s) — both green, reinforcing that nothing doctest-adjacent broke.

## Not checked

- Full `Pkg.test()` / Aqua / package load — excluded per the concurrent-review constraint; relying on the green CI matrix (`gh pr checks 14`) instead, which the PR body also cites (30 testsets, 0 failed).
- Type piracy / inference / allocations — not applicable rather than unmeasured: the parse trees are identical apart from the two printed labels, so there is no executable change to attribute anything to.

## Findings

None. Every one of the four recurring defect classes from `Tasks/Normalise the Julia sources to NFC.md` was checked against this PR specifically (not the tree-wide history) and none reproduces here: the glyph set and counts are exactly right, the "one runtime-visible change" is honestly and correctly scoped rather than denied, the counts are correct, and the CHANGELOG heading matches its file's own shape.

PR #15, opened 2026-09-19
Rewrite on Symbolics.jl; fields become values

michakraus:
Rewrites the symbolic machinery on Symbolics.jl and makes an electromagnetic field a value
instead of a set of names injected into a module. Breaking, targeting v0.9.0.

## What changes for a caller

A field is built by `FieldFunctions` and passed around like any other object:

```julia
field = FieldFunctions(AxisymmetricTokamakCylindricalEquilibrium(6.2, 5.3, 2.0))

B♭(field, t, ξ)      # covariant components of B, an SVector{3}
B(field, t, ξ)       # |B|, a scalar
```

`@code`, `code` and `load_equilibrium` are gone, and with them the world-age trap that the
do-block form of `load_equilibrium` existed to work around. A field can now be constructed
inside a function, stored, and handed to a vector field that depends on it — typically through a
`GeometricEquations` problem's `parameters`.

The ~450 scalar functions per equilibrium become ~41 tensor-valued generics returning
StaticArrays, named with the musical isomorphisms: `♭` lowers an index (covariant), `♯` raises
one (contravariant), `♮` gives the physical components, and the bare letter is the magnitude, so
`B` is still `|B|`. `B♭(field, t, ξ)[1]` is the old `B₁(t, ξ)`.

The per-field submodules are flattened into the package: `ThetaPinch.init()` is now
`ThetaPinchEquilibrium()`, and the six Solov'ev alias modules are replaced by the constructors
that already existed inside `Solovev`.

`CHANGELOG.md` has the full list, including the name-by-name mapping.

## Nothing generated was lost

`scripts/verify_against_symengine.jl` evaluates every one of the ~444 functions that v0.8
generated, for each of the 20 equilibria, and compares against values recorded from that
implementation in `scripts/reference-symengine.tsv`.

**8844 of 8844 values agree. Zero unresolved names, zero frame differences** — so Symbolics even
selects the same trial vector for the perpendicular frame that SymEngine did.

## Parameters are arguments, not literals

The trace runs against a copy of the equilibrium whose parameters are symbolic, so the generated
code is written in terms of `R₀`, `B₀`, `q₀` … and takes their values as an argument. The
equilibrium struct stays where the values live.

The code therefore depends on the equilibrium's *type* and not on its parameters, which makes two
things sound: the generated functions are cached, keyed on types rather than values, and a
`PrecompileTools` workload traces one field of every shipped type during precompilation, so the
specializations and the cache both land in the package image.

| | before | after |
|:--|--:|--:|
| first field built in a session | 8.5 s | 0.00 s |
| all twenty shipped equilibria | ~19 s | **0.01 s** |
| package precompilation | 1.5 s | 24.5 s |

The precompilation cost is paid once per version. A downstream package can go further and build
its own fields during *its* precompilation with `@precompilable_fields` and `cache_module`,
bringing the first evaluation in a session to microseconds; this was measured end to end on a
throwaway package.

The cache is keyed on types, not on the content of the methods behind them, so redefining an `A₁`
mid-session leaves it stale: `clear_field_cache!()` discards it and `cache = false` bypasses it.

## Dependencies

`SymEngine` out. `Symbolics`, `StaticArrays`, `PrecompileTools`, `RuntimeGeneratedFunctions` and
`GeometricBase` in. The Julia floor stays 1.10, so the `min` CI job does not move.

`GeometricBase` is there for the generic names `functions`, `parameters` and `periodicity`, which
this package now extends rather than defining for itself — it previously exported a `periodicity`
of its own, which collided with `GeometricEquations`' in any package using both.

## Verification

- `Pkg.test()` passes, including new coverage for the symbolic parameters and the cache.
- The documentation builds clean, through rendering.
- Every one of the 41 accessors is allocation-free and type-stable, asserted in the suite for all
  20 equilibria and checked independently across containers and element types.
- `Aqua` piracy, ambiguities, unbound args, stale deps, undefined exports, project extras and
  persistent tasks all pass. `ExplicitImports` finds no stale import.
- JuliaFormatter and the NFC check are clean on every changed file.

## Open points for the reviewer

Two things I could not settle without a decision:

- **`src/analytic/solovev.jl` uses `Symbolics.symbolic_to_float`, which is not public in
  Symbolics** (`Base.ispublic` is `false`). It is needed because `substitute` folds arithmetic
  but not `NaNMath.log`, so a fully substituted expression is still symbolic. `[compat] Symbolics
  = "7"` allows any 7.x to remove it. A public-API replacement via `build_function` was measured
  to agree to 1 ULP and to be no slower, at the cost of one generated function per basis
  expression; it is not applied here.
- **`periodicity(field)` returns four ±Inf components for every equilibrium**, so the toroidal
  chart reports no periodicity in ϕ although its own `rangemax` is 2π. The default is inherited
  unchanged from v0.8 and no equilibrium overrides it; what is new is that the struct stores it
  and the docstring calls it "the periodic domain".

Pre-existing, unchanged from `main`, but now rendered on the new Penning Traps documentation
page:

- `PenningTrapBottle`'s `A` and `B` docstring formulas both disagree with the code — at
  `(0.3, 0.4, 0.5)` the code gives `A♭ = [-37.87, 14.10, -12.00]` where the docstring gives
  `[-50, -25, 25]` — and it names the parameter `B₁` where the field is `Bₚ`.
- `PenningTrapAsymmetric`'s docstring gives the `B₀` coefficient of `B₁` as `1/3`; the code gives
  `1/6`.
- `contravariant_to_covariant`, `contravariant_to_physical`, `normalize(v, g)`,
  `to_cartesian(ξ, equ)` and `from_cartesian(x, equ)` have no caller, and
  `contravariant_to_physical` carries `covariant_to_physical`'s docstring.
- `LinearAlgebra` has no `[compat]` entry, and `AxisAspect` is an unused import in the Makie
  extension.

## Follow-on

`ChargedParticleDynamics` is the only package in the ecosystem that depends on this one, and it
uses the `@code` macro throughout. Porting it is a separate pull request against this version.


michakraus (commented):
**Verdict: request changes.**

The rewrite itself holds up under every check I could run against it, including the two headline
claims, which reproduce exactly. What blocks it is narrow: the package does not load at all on the
Julia version its own `[compat]` declares as the floor.

## CI

`Julia min` is red on ubuntu, macOS **and** Windows. Read from the job rather than from the
workflow conclusion, it is not a test failure — the package fails to *precompile*, so
`using ElectromagneticFields` never succeeds:

```
ERROR: LoadError: OverflowError: 1474560000 * -33397145600 overflowed for type Int64
  [ 4] _mul(::Type{Rational{Int64}}, ...)   DynamicPolynomials/src/mult.jl:147
  ...                                       SymbolicUtils/src/polyform.jl:290
  ...                                       Symbolics/src/Symbolics.jl:278    (simplify)
  [69] generate_field_expressions(equilibrium::SolovevXpointEquilibrium{Float64}, ...)
                                            src/analytic/analytic_field.jl:457
  [77] macro expansion                      src/ElectromagneticFields.jl:124  (@compile_workload)
```

`min` is a required check, so this cannot merge as it stands.

Everything else is green: `Julia 1` on all three platforms, `pre`, `nightly`, `Documentation` and
`Doctests`. One side effect worth knowing: `CI.yml` sets `COVERAGE` only on the `min` + ubuntu
matrix entry, which is the failing one, so no coverage was uploaded for this pull request.

## Findings

| 1 | blocker | `src/analytic/analytic_field.jl:457`, `Project.toml:32` | The package cannot be loaded on Julia 1.10, the declared minimum. `Symbolics.simplify.(avec)` in the perpendicular-frame trial loop overflows `Int64` inside the rational polynomial gcd while the precompile workload traces `SolovevXpointEquilibrium`. | Reproduced locally, independently of CI: `julia +1.10 --project=<worktree> -e "using ElectromagneticFields"` gives `OverflowError: 64715489280 * -175841280 overflowed for type Int64`, same frame at `analytic_field.jl:457`, Julia 1.10.12 / macOS aarch64 / Symbolics v7.40.0. I then checked the two versions CI does not cover: **1.11 and 1.12 both load fine** (`LOADED OK 1.11`, `LOADED OK 1.12`), and 1.10 and 1.13 resolve the same Symbolics v7.40.0, so this is a Julia-version difference and not a resolver one. Either raise the floor to 1.11, or take the `simplify` out of the frame selection (`iszero` on an expanded form would do the same job). |
| 2 | should-fix | `src/analytic/analytic_field.jl:140-142`, `field_functions.jl:303` | `periodicity(field)` returns **four** components for a three-coordinate chart. | `periodicity(FieldFunctions(AxisymmetricTokamakCylindricalEquilibrium()))` gives `([-Inf, -Inf, -Inf, -Inf], [Inf, Inf, Inf, Inf])`, a `Tuple{Vector{Float64}, Vector{Float64}}` of length 4 each, while `rangemax` for the same field is `[Inf, Inf, 6.283185307179586]`. The CHANGELOG justifies adding GeometricBase precisely so this generic is shared with GeometricEquations, which destructures it as `per_lo, per_hi = periodicity(equ)` — so what a downstream `ODEProblem(...; periodicity = periodicity(field))` receives is a 4-vector for a 3-dimensional state, reporting no periodicity in the toroidal angle. It is exported, documented in `docs/src/interface.md:136`, and `grep periodicity test/` returns nothing. You flagged this as an open point; v0.9.0 is the release where the generic changes owner, so it is the right place to settle it. |
| 3 | should-fix | `Project.toml:3` vs `CHANGELOG.md:10` | The version is bumped to `0.9.0` while the CHANGELOG still reads `## [Unreleased] — targeting 0.9.0`. | `Packages/CLAUDE.md`, *Releases*: close out the Unreleased heading and set `version` **in the same commit**, and that commit is the release. Merging this leaves `main` carrying a version that was never registered, and the later release commit with nothing to bump. Leave `version = "0.8.0"` here. |
| 4 | should-fix | `docs/src/interface.md:166-172`, `test/test_analytic.jl:32` | The documentation states that type stability is asserted by the test suite; it is not. | `grep -rn "@inferred" test/` returns nothing. What `test_interface` asserts is `(v isa Number ? typeof(v) : eltype(v)) === eltype(ξ)` for one `Vector{Float64}` input — a value-level check on the element type, not an inference check, and not "across containers and element types" as the PR body says. I measured inference separately and it **is** clean, so this is a claim-versus-coverage gap rather than a defect; a `@inferred` line in `test_interface` and one non-`Float64` container would close it. |
| 5 | should-fix | `src/analytic/analytic_field.jl:322`, `src/analytic/field_functions.jl:284` | `Base.typename(...).wrapper` is a non-public Base internal, and it carries both the symbolic-copy reconstruction and the cache key. | `ExplicitImports.print_explicit_imports(...; report_non_public = true)`: *"typename is not public in Base but it was accessed via Base at src/analytic/analytic_field.jl:322:20"*. The PR body raises `Symbolics.symbolic_to_float` as the one non-public dependency; this is the same kind of exposure on a more load-bearing path, and it is not mentioned. `Base.typename(T).wrapper` has no public spelling, so this may end as a documented accepted risk — but it should be stated alongside the other one. |
| 6 | nit | `src/analytic/field_functions.jl:284-285`, `:294` | The cache key conflates the equilibrium and perturbation parameter counts, and the assertion that guards it cannot catch the collision. | The key holds `length(pvalues)`, which is the **sum** over both structs, and `@assert generated.nparameters == length(pvalues)` compares that same sum. Two `(equ, pert)` pairs of the same wrapper types whose vector parameters split differently but total the same therefore share one cache entry and the assertion passes. Unreachable with the shipped perturbations, which all have fixed scalar parameters; reachable for a user perturbation with a vector parameter. Storing the two lengths separately costs nothing. |
| 7 | nit | `docs/src/interface.md:41`, `docs/src/generation.md:115`, `:309-311` | Three reference passages still narrate the migration, which is what commit `0e005cd` set out to remove from the prose. | `interface.md:41` — *"The `x` of the old component names — `dA₁dx₂`"*; `generation.md:115` — *"There is no equivalent of printing the generated module, because there is no module"*; `generation.md:309-311` — *"values recorded from the SymEngine-based implementation this one replaced"*. A migration guide is the right home for all three; the Interface and Code Generation reference pages describe the package as it is. |
| 8 | nit | `src/analytic/analytic_field.jl:144-145`, `field_functions.jl:308-332` | Four exported names carry two unrelated calling conventions, and the generated docstrings document only one. | `methods(B)` returns five entries: three `B(x::AbstractVector, equ)` and two `B(field::FieldFunctions, t, ξ)`. Same for `J`, `to_cartesian` and `from_cartesian`. The two conventions differ in return type as well — `to_cartesian(ξ, equ)` builds a heap `Vector`, `to_cartesian(field, t, ξ)` an `SVector`. No ambiguity, since the arities differ, but the docstring a user sees for `to_cartesian` mentions only the field form. |
| 9 | nit | `CHANGELOG.md:12` | The Unreleased block has only `### Changed`. | The removals are substantial and load-bearing for a breaking release — `@code`, `code`, `load_equilibrium`, SymEngine, the per-field submodules, the six Solovev alias modules — and Keep a Changelog, which the file header cites, puts them under `### Removed`. They are all described, just under the wrong heading. |

## What I reproduced

Both headline claims hold, measured here rather than taken from the PR body.

**The SymEngine reference.** `julia +1.11 scripts/verify_against_symengine.jl`:

```
total 8844 reference values: 8844 checked, 0 unresolved, 0 frame differences, 0 FAILED
```

Every one of the twenty equilibria reports `0 FAILED`, `0 frame differences`, `0 unresolved`.
This is the strongest single piece of evidence in the pull request, and it is exactly as claimed.

**Allocations, inference and element type.** Cold Julia 1.11 process, `--check-bounds=auto`, the
measurement behind a `@noinline` function barrier — without the barrier every call appears to
allocate 16 to 672 bytes, which is the boxing of the dynamically dispatched return value and not
the call, exactly as `docs/src/interface.md:174-176` warns. All 41 accessors, four coordinate
containers (`Vector{Float64}`, `SVector{3,Float64}`, `Vector{Float32}`, `Vector{Int}`), four
equilibria:

```
AxisymmetricTokamakCylindricalEquilibrium     noninferred=0 allocating=0 wrongtype=0
SolovevEquilibrium                            noninferred=0 allocating=0 wrongtype=0
ThetaPinchEquilibrium                         noninferred=0 allocating=0 wrongtype=0
AxisymmetricTokamakToroidalEquilibrium        noninferred=0 allocating=0 wrongtype=0
```

`Test.@inferred` passes for every combination, `@allocated` is 0 for every combination, and every
return is `float(eltype(ξ))`. So the property is real; finding 4 is only that the suite does not
assert the inference half of it.

**Precompilation cost.** 27.7 s for ElectromagneticFields on Julia 1.11, aarch64 macOS, against
the 24.5 s in the PR body. Consistent.

## Verified and fine

- **JuliaFormatter**, sciml style, all 26 changed `.jl` files: clean.
- **Aqua** on Julia 1.11: piracy, method ambiguity, unbound type parameters, undefined exports,
  stale dependencies and the Project/test-Project comparison all pass. The only failure is
  `deps_compat`, on `LinearAlgebra` and on the `Test`/`SafeTestsets` extras — all three
  pre-existing.
- **ExplicitImports**: no stale explicit import. The non-public accesses it does report are
  finding 5 and the two the PR body already raises.
- **No type piracy** in the new code. `functions`, `parameters` and `periodicity` are GeometricBase
  generics extended on types this package owns, which is the point of taking the dependency.
- **CHANGELOG**: present, thorough, with the name-by-name mapping, and the existing `[Unreleased]`
  entry from the NFC pull request is preserved rather than overwritten — the heading is the only
  line the diff removes.
- **File list taken from git**, not from `gh pr view --json files`: 54 files, so no truncation
  either way.
- No `close #NN` in the description, so nothing is auto-closed on merge.
- `Julia 1.11` and `Julia 1.12` both load the package, which CI does not cover.

## Pre-existing, reported and not fixed

Beyond the ones you already list in the description:

- `src/analytic/solovev.jl:109` and `:272` — the docstring gives
  `\psi_{5} = 2 y^4 9 y^2 x^2 + ...`, missing the minus sign between the first two terms. The code
  at `solovev.jl:33` has `2 * x[2]^4 - 9 * x[2]^2 * x[1]^2`. Byte-identical on `main`.
- `SafeTestsets` sits in `[extras]` and `[targets]` but `test/runtests.jl` never uses it.
- `normalize` and `normalize!` at `analytic_field.jl:244` and `:249` shadow `LinearAlgebra.normalize`
  inside the module, which is `using LinearAlgebra`.

Since finding 3 will touch `Project.toml` anyway, the missing `LinearAlgebra` `[compat]` entry
could go in the same change rather than into a follow-up nobody opens.

## Not checked

- The full `Pkg.test()` locally. CI runs it green on 1.13 across three platforms, and another
  session on this machine is holding long Julia jobs on 1.13; running a suite with
  `--check-bounds=yes` beside them over shared packages invalidates precompile images and produces
  failures that are not real.
- The documentation build locally — `Documentation` and `Doctests` are both green in CI.
- The Makie extension output. `test/test_plots.jl` exercises it and CI is green.
- The "before" column of the timing tables, which needs a 0.8 checkout with SymEngine. The "after"
  column and the precompilation figure I did measure, and both hold.

michakraus (commented):
## Review — Rewrite on Symbolics.jl; fields become values

**Verdict: request changes.** The rewrite itself reads well and the SymEngine cross-check is a
strong piece of evidence. But the package does not load on Julia 1.10, which `[compat]` still
declares as the floor, and the allocation guarantee the PR advertises does not hold there either.
Both are new in this branch: the last CI run on `main` (run 35268028697) is green across the
whole matrix, `min` included.

### CI, read per job rather than per workflow

`CI.yml:48` sets `continue-on-error: ${{ matrix.experimental }}`, so `pre` and `nightly` would
report green even on failure. I checked every job's step conclusions rather than trusting the
badges:

```
$ gh api repos/JuliaPlasma/ElectromagneticFields.jl/actions/runs/35428761148/jobs --paginate \
    --jq '.jobs[] | [.name, .conclusion, ([.steps[] | select(.conclusion=="failure") | .name] | join(";"))] | @tsv'
Doctests - ubuntu-latest                success
Julia min - macOS-latest - default      failure  Run julia-actions/julia-runtest@v1
Julia 1 - windows-latest - default      success
Julia 1 - ubuntu-latest - default       success
Julia 1 - macOS-latest - default        success
Julia min - windows-latest - default    failure  Run julia-actions/julia-runtest@v1
Julia pre - ubuntu-latest - default     success
Julia min - ubuntu-latest - default     failure  Run julia-actions/julia-runtest@v1
Julia nightly - ubuntu-latest - default success
```

`pre` and `nightly` are genuine passes — no failed step, so `continue-on-error` is masking
nothing. `min` is Julia 1.10.12 and `Julia 1` is Julia 1.13.0. The two failure modes differ by
platform, and both are real.

### Findings

| 1 | blocker | `src/analytic/analytic_field.jl:457`, `src/ElectromagneticFields.jl:103` | The package does not precompile on Julia 1.10. `Symbolics.simplify` overflows `Int64` in the zero test for the perpendicular frame; because the trace runs inside `@compile_workload`, this aborts precompilation and the package cannot be loaded at all. | See "The 1.10 precompile failure" below — verbatim CI errors, a local reproduction, and a one-line change that clears it. |
| 2 | blocker | `test/test_analytic.jl:35` | All 41 accessors allocate on Julia 1.10, so `min` stays red even once finding 1 is fixed. `--check-bounds` is not the cause. | See "The 1.10 allocation failure" below. |
| 3 | correctness-risk | `Project.toml:31`, `src/analytic/solovev.jl:74` | `Symbolics = "7"` is an open bound on a non-public name. Confirmed non-public on the resolved version: `Base.ispublic(Symbolics, :symbolic_to_float)` is `false` and `Base.isexported` is `false`, on Symbolics v7.40.0, where `substitute`, `build_function` and `simplify` are all `true`. Any 7.x may remove it. | The PR body raises this; I verified it. Either take the `build_function` route already measured to agree to 1 ULP, or cap the bound. |
| 4 | quality | `CHANGELOG.md:54`, `docs/src/generation.md:174`, `src/ElectromagneticFields.jl:98` | "all twenty shipped equilibria — 0.01 s" measures only `FieldFunctions` on already-built structs. It excludes the constructors, which dominate. | See "Re-measuring the timing claim" below: 0.014 s for the twenty `FieldFunctions` calls, but 0.49 s to construct the twenty structs first. |
| 5 | quality | `test/runtests.jl:1` | The suite is `test_analytic.jl` and `test_plots.jl` only. There is no Aqua and no ExplicitImports test, so the PR body's "Aqua … all pass. `ExplicitImports` finds no stale import" is a one-off local result that nothing re-checks. Related: `LinearAlgebra` (`Project.toml:9`) has no `[compat]` entry, which an in-suite `Aqua.test_deps_compat` would hold you to. | `cat test/runtests.jl` shows four lines: `using ElectromagneticFields`, `using Test`, and the two `include`s. |
| 6 | quality | `src/analytic/analytic_field.jl:140` | `periodicity` returns two length-**4** vectors for a **3**-coordinate chart, while `rangemin`/`rangemax` are documented as `SVector{3}` (`src/analytic/field_functions.jl:54`). Pre-existing — `git show origin/main:src/analytic/analytic_equilibrium.jl` line 102 is identical — but this branch newly extends `GeometricBase.periodicity`, stores the value in the struct (`field_functions.jl:186`) and documents it as "the periodic domain" (`field_functions.jl:157`), so it is now advertised interface. The arity should be settled in this PR rather than inherited. | Read at both revisions; `ones(T, 4)` against `SVector{3}` bounds. |
| 7 | quality | `src/analytic/penning_trap_bottle.jl:12` | The docstring's `A` and `B` formulas are interchanged, and the new `docs/src/analytic/penning_traps.md` page now renders them. The code's `A₁` (line 59) is `-B₀/2·y - Bₚ(yz² - y³/6)`, which is the shape the docstring gives for **`B`** (line 16), not for `A` (line 12). The docstring also writes the parameter as `B_1` where the struct field is `Bₚ`. Numerically: `A♭(f, 0.0, [0.3, 0.4, 0.5])` returns `[-37.86666666666667, 14.1, -12.0]`, against `[-50, -25, 25]` from the docstring. | Ran against the PR head on Julia 1.13, and read lines 5-32 and 58-65. The PR body raises it as pre-existing; publishing it on a new page is what makes it this PR's. |
| 8 | nit | `ext/ElectromagneticFieldsMakieExt.jl:4` | `AxisAspect` is imported and never used — one occurrence in the file, the import itself. Pre-existing (identical on `main`), and the file is touched here anyway. | `grep -c AxisAspect ext/ElectromagneticFieldsMakieExt.jl` → `1`. |
| 9 | nit | `Project.toml:36` | `SafeTestsets` is a test dependency with no use anywhere. Pre-existing on `main`. | `grep -rn "SafeTestsets\|safetestset" test/ src/` → no matches. |

### The 1.10 precompile failure

`Julia min - ubuntu-latest`, verbatim:

```
ERROR: The following 2 direct dependencies failed to precompile:
Failed to precompile ElectromagneticFields [d6c1ba6f-ee03-53af-b876-68cefeb88ec8] to "/home/runner/.julia/compiled/v1.10/ElectromagneticFields/jl_XWmQ09".
ERROR: LoadError: OverflowError: 1474560000 * -33397145600 overflowed for type Int64
```

The frames that matter:

```
 [1] throw_overflowerr_binaryop(op::Symbol, x::Int64, y::Int64)  @ Base.Checked ./checked.jl:163
 [4] _mul(::Type{Rational{Int64}}, …)                            @ DynamicPolynomials … mult.jl:147
[19] primitive_univariate_gcd!(…)                                @ MultivariatePolynomials … gcd.jl:620
[35] safe_gcd(…)                                                 @ SymbolicUtils … polyform.jl:290
[36] simplify_div(…)                                             @ SymbolicUtils … polyform.jl:299
[64] generate_field_expressions(equilibrium::…SolovevXpointEquilibrium{Float64}, perturbation::…ZeroPerturbation)
     @ ElectromagneticFields …/src/analytic/analytic_field.jl:457
[69] ElectromagneticFields.FieldFunctions(equ::…SolovevXpointEquilibrium{Float64})
[76] top-level scope  @ …/src/ElectromagneticFields.jl:103
```

`Julia min - windows-latest` fails the same way but on a **different equilibrium and different
operands** — `OverflowError: 4505833835455315968 * 237049338986496 overflowed for type Int64`,
under `SolovevEquilibrium{Float64}`. `Julia min - macOS-latest` does not overflow at all. I
reproduced it on Julia 1.10.12 locally with a third operand pair,
`OverflowError: 64715489280 * -175841280 overflowed for type Int64`, again at
`analytic_field.jl:457` under `SolovevXpointEquilibrium{Float64}`, and it repeats identically
across three runs on the same machine. So it is deterministic per platform and varies between
platforms.

Two things make this a blocker rather than an upstream annoyance, and both are in your control:

- **It is inside `@compile_workload`.** `src/ElectromagneticFields.jl:103` runs the trace during
  precompilation, so one bad equilibrium takes the whole package down. With the workload disabled
  through `PrecompileTools`' `precompile_workload = false` preference, the package loads fine on
  1.10 and every equilibrium except the offending Solov'ev one traces correctly. As it stands,
  `using ElectromagneticFields` fails on 1.10.
- **The `simplify` at line 457 is only a zero test.** It asks whether a trial cross product
  vanishes; it does not need full simplification, and full simplification is what drags in the
  `Rational{Int64}` polynomial gcd that overflows.

I tested the obvious fix. Overriding `Symbolics.simplify(::Num)` with `Symbolics.expand` in a
throwaway process on Julia 1.10, so line 457 uses the cheaper test, every equilibrium traces:

```
SolovevXpointITER          ok   B♭ = [-0.024306583927461305, 0.01553587171840426, 5.300000000000001]
SolovevITER                ok   B♭ = [-0.023281988077396246, 0.015853814608476232, 5.300000000000001]
SolovevDoubleXpointNSTX    ok   B♭ = [-0.13607184066508146, -0.17155247684776792, 0.30000000000000004]
ABC                        ok   B♭ = [1.3570081004945758, 1.7450057874843896, 0.97699658649593]
AxiTorRegularization       ok   B♭ = [0.0, 0.28689096580652335, 1.0]
julia = 1.10.12
```

And it picks the same trial vector, so the frame is unchanged. `a♭` and `c♭` at
`ξ = [1.05, 0.5, 0.5]`, `expand` on 1.10 against `simplify` on 1.13, agree to all 12 digits printed:

```
ABC            a♭ = [0.0, -0.488524703096, 0.87255006416]      c♭ = [0.827487576091, -0.489922922881, -0.274298817078]
SolovevITER    a♭ = [0.0, 6.199969418915, -0.020446833063]     c♭ = [6.199934049595, 8.9818021e-5, 0.030026559764]
ThetaPinch     a♭ = [0.0, -1.0, 0.0]                           c♭ = [1.0, 0.0, 0.0]
AxiCyl         a♭ = [0.0, 0.999687646408, -0.026241800718]     c♭ = [0.970160322999, 0.006059714697, 0.254508017277]
```

I am not claiming `expand` is the right zero test in general — it is a probe, and you know the
expressions better than I do. But it localises the fault to that one line and shows the fix does
not need a compat bump or a floor bump. Separately, please consider whether the workload should
tolerate a trace that throws; turning one equilibrium's upstream arithmetic bug into "this package
will not load" is a lot of blast radius for a precompilation optimisation.

### The 1.10 allocation failure

`Julia min - macOS-latest` gets past precompilation and then fails, verbatim:

```
ABC : Test Failed at /Users/runner/work/ElectromagneticFields.jl/ElectromagneticFields.jl/test/test_analytic.jl:35
  Expression: #= …test/test_analytic.jl:35 =# @allocated(f(field, t, ξ)) == 0
   Evaluated: 1680 == 0
…
Test Summary: | Pass  Fail  Total   Time
ABC           |  145    41    186    8.3s
ERROR: LoadError: Some tests did not pass: 145 passed, 41 failed, 0 errored, 0 broken.
```

41 failures is all 41 accessors, for the first equilibrium only — the suite aborts there, so the
other nineteen never ran on that job. Observed values range from 1664 to 2320 bytes.

I checked the obvious explanation first and it is **wrong**. `Pkg.test()` forces
`--check-bounds=yes` on 1.10 but inherits the caller's setting from 1.13 on, which would explain a
min-only allocation failure. It does not:

| julia | `--check-bounds` | accessors allocating | max |
|:--|:--|--:|--:|
| 1.13.0 | `auto` | 0 / 41 | 0 B |
| 1.13.0 | `yes` | 0 / 41 | 0 B |
| 1.10.12 | `auto` | **41 / 41** | 672 B |

Fresh process each row, same worktree, same resolved dependencies, one warm-up call before each
`@allocated` exactly as `test_analytic.jl:34` does, `ABCEquilibrium`. So this is genuinely Julia
1.10 and not the test harness. Inference is not the cause either — on 1.10,
`Base.return_types(B♭, Tuple{typeof(field), Float64, Vector{Float64}})` is
`[SVector{3, Float64}]` and `@inferred B♭(field, t, ξ)` passes. Type-stable, and still allocating;
my guess is the `RuntimeGeneratedFunction` call in `FieldFunction` (`field_functions.jl:33`) not
inlining on 1.10, but I did not confirm that.

This needs a decision, and I do not think it is mine to make:

- raise `[compat] julia` to `"1.11"` or `"1.12"` and move the `min` job with it — honest, and the
  PR body's "the Julia floor stays 1.10, so the `min` CI job does not move" becomes false rather
  than the CI being false; or
- find what allocates on 1.10 and fix it; or
- keep the floor and scope the allocation assertion to the versions where it holds — but then
  please also soften the claim in `CHANGELOG.md` and `docs/src/generation.md`, because
  "every one of the 41 accessors is allocation-free" is not true on the supported floor.

Note that dependency versions are *identical* between the failing and passing jobs — Symbolics
v7.40.0, SymbolicUtils v4.46.7, DynamicPolynomials v0.6.8, MultivariatePolynomials v0.5.19 in both
`Julia min - ubuntu` and `Julia 1 - ubuntu`. So neither failure is a resolver artefact.

### Re-measuring the timing claim

Cold process, Julia 1.13, PR head, repeated twice with agreeing results:

| what | measured |
|:--|--:|
| `using ElectromagneticFields` | 0.94 s |
| first field (cylindrical tokamak) | 0.001 s |
| `FieldFunctions` on all twenty, structs already built | **0.014 s** |
| constructing the twenty equilibrium structs | **0.491 s** |
| twenty constructors + `FieldFunctions`, as a user would write it | 0.68 s |

The 0.01 s is right for the step it measures, and the cache clearly works. But
`docs/src/generation.md:174` presents it as "all twenty shipped equilibria", which a reader will
take as wall time from a fresh session, and that is ~0.5 s. The difference is the Solov'ev
constructors: `SolovevEquilibriumITER()` costs 0.30 s on its first call and 2.3 ms warm, because
the coefficient solve runs `Symbolics.@variables` and `_solovev_eval` per construction and is not
covered by the cache. Worth a clause saying what the figure excludes, in all three places it
appears. I did not re-measure the "before" column — that needs the SymEngine implementation.

### Verified and fine

- **The precompilation cost.** 22.1 s to precompile the package here, against 24.5 s claimed.
  Close enough, and it is a real cost that the runtime figures do buy back.
- **Every shipped type is in the workload.** I went looking for `SolovevDoubleXpointEquilibrium`
  as a gap and it is not one — `solovev.jl:462` returns a `SolovevXpointEquilibrium{T}`, which
  `SolovevXpointEquilibriumITER()` already covers. The claim holds.
- **Formatting and NFC**, on all 26 changed `.jl` files and all 54 changed files respectively.
  JuliaFormatter with the repo's `style = "sciml"` reports every file already formatted; the NFC
  check exits 0.
- **`Julia 1` is a real pass** on all three platforms — test summaries show Pass columns only, no
  Fail column.
- **The cache key** (`field_functions.jl:284`) drops type parameters deliberately and carries
  `length(pvalues)` to cover the Solov'ev coefficient vector. The `@assert` on `nparameters`
  cross-checks two independent paths rather than restating the key, so it is doing real work.
- No `close #NN` in the PR body, so nothing is auto-closed on merge.

### Not checked

- **The 8844-value SymEngine cross-check.** I did not re-run `scripts/verify_against_symengine.jl`;
  it needs the old implementation, and the recorded `.tsv` is what the script compares against, so
  re-running it here would only confirm the file matches itself.
- **The full test suite.** I read the existing CI logs rather than re-running it.
- **Whether the other nineteen equilibria also allocate on 1.10.** The macOS min job aborted after
  `ABC`, and I measured `ABC` locally. The pattern is likely general, but I only have the one.
- **Why `RuntimeGeneratedFunction` allocates on 1.10.** Stated as a guess above, not confirmed.
- **The `PenningTrapAsymmetric` `1/3` vs `1/6` coefficient**, and the uncalled helpers
  (`contravariant_to_covariant` and friends) listed in the PR body. I verified the
  `PenningTrapBottle` discrepancy and took the rest on trust.
- **Documentation rendering.** The `Documentation` and `Doctests` jobs pass; I did not build docs.

michakraus:
Pushed `80321e6`, which fixes the two blockers. Both were reproduced locally on Julia 1.10.12
before being fixed, and both fixes were verified there afterwards.

Two reviews are on this pull request and their numbering differs. The table below uses the numbers
from the later review; the earlier review's distinct findings follow underneath.

## Blockers — fixed

**1 — the package did not precompile on Julia 1.10.** Confirmed as stated.

```
ERROR: LoadError: OverflowError: 1474560000 * -33397145600 overflowed for type Int64
```

reproduced locally as `OverflowError: 64715489280 * -175841280`, same site, on 1.10.12.

`Symbolics.simplify` is used at `src/analytic/analytic_field.jl:457` only as a zero test. It runs
`simplify_fractions`, whose polynomial gcd over `Rational{Int64}` is what overflows. The zero test
does not need fraction cancellation, so the call now passes `simplify_fractions = false` — a
documented keyword of `SymbolicUtils.simplify`, present in every SymbolicUtils version the
`Symbolics = "7"` bound can resolve.

This keeps the floor at 1.10 and needs no compat change.

*The frame is unchanged.* A wrong trial vector would still be orthonormal, so the suite's
orthonormality assertions cannot settle it on their own. Comparing against the unmodified branch on
1.13, where the original call works, `a♭` and `c♭` agree to all twelve digits printed for all
fifteen precompiled equilibria — no line of the comparison differs.

**2 — all 41 accessors failed the allocation assertion on 1.10.** Confirmed, and it is a real
second blocker: the macOS `min` job got past precompilation and reported

```
ABC                                                          |  145    41   186
  Expression: @allocated(f(field, t, ξ)) == 0
   Evaluated: 1680 == 0
```

with 1664, 1680, 1728, 1872 and 2320 across the 41.

The cause is measurement, not the package. `f` arrives from a lazy generator, so the call is
dynamically dispatched and `@allocated` charges for boxing the result. 1.13 elides the box and 1.10
does not, which is why only `min` went red. Measured behind a `@noinline` barrier, **1.10 and 1.13
both report 0 allocations for all 41 generics**, so the assertion stays `== 0` — no tolerance was
widened. `docs/src/interface.md` already prescribes exactly this barrier for users; the test now
uses it too.

On `main` these assertions called each generic by name, which is statically dispatched. The
generator loop is what introduced the dispatch, so the finding is attributable to this branch.

Verification after both fixes, on **Julia 1.10.12 with `--check-bounds=yes`**, which is what
`Pkg.test()` forces there: all 20 equilibrium testsets and the five remaining testsets pass,
3839 assertions, 0 failures. Unchanged on 1.13.

## Not fixed — pre-existing, and so reported rather than changed here

| # | file | why it was left |
|--:|:-----|:----------------|
| 6 | `src/analytic/analytic_field.jl:140` | `periodicity` returning length-4 vectors for a 3-coordinate chart is **byte-identical** at `origin/main:src/analytic/analytic_equilibrium.jl:102`. Real, and worth its own change. |
| 7 | `src/analytic/penning_trap_bottle.jl` | The interchanged `A`/`B` docstring formulas are untouched by this diff, which only removes the module wrapper and renames the constants. |
| 8 | `ext/ElectromagneticFieldsMakieExt.jl:4` | The `AxisAspect` import line is byte-identical on `main`. It is genuinely unused **inside the extension** — the two remaining uses are in `docs/`, which reach Makie directly. |
| 9 | `Project.toml` | `SafeTestsets` in `[extras]`/`[targets]` is unused on `main` as well. |

## Not fixed — these need a decision, not a fix

- **3, `Symbolics = "7"` over a non-public name.** Tightening the bound is a judgement about how
  much upstream movement to absorb, and the body already accepts `symbolic_to_float`.
- **5, no Aqua or ExplicitImports in the suite.** Adding either means adding a test dependency.
  The `LinearAlgebra` `[compat]` gap is real and is a one-line addition whenever `Project.toml` is
  next touched.
- **4, the "all twenty — 0.01 s" figure.** Not re-measured here, so it is neither confirmed nor
  disputed. It is a documentation claim and does not gate CI.

From the earlier review, also left for a decision:

- **The `version = "0.9.0"` bump while `## [Unreleased] — targeting 0.9.0` is still open.** Two
  fixes are valid — revert the bump, or close the heading out — and the second belongs to the
  release commit. Worth settling before merge, since it decides whether this pull request is the
  release.
- **`docs/src/interface.md` stating that type stability is asserted by the suite.** `@inferred`
  appears nowhere in `test/`. The property itself holds; the claim is ahead of the coverage. It can
  be closed by softening the sentence or by adding the assertion.
- **`Base.typename(...).wrapper`** on the symbolic-copy and cache-key paths, which is a second
  non-public dependency and is not listed beside `symbolic_to_float`.
- **The three passages that still narrate the migration**, and the `### Removed` heading the
  CHANGELOG is missing.

## Disputed

- **The cache key at `field_functions.jl:284`.** The two reviews disagree: one calls the shared
  `length(pvalues)` a latent collision, the other calls the `@assert` an independent cross-check.
  Unreachable with the shipped perturbations either way, so nothing was changed.
- **One figure.** A "max 672 B" allocation magnitude was quoted in review. Both CI and local
  measurement give 1664–2320 B. The finding stands; that number does not.

No CHANGELOG entry was added. Both fixes repair code that this same unreleased block introduces, so
there is nothing for a reader of 0.9.0 to be told.

michakraus:
Correction to my earlier comment, and two further commits.

## The first fix was wrong, and CI caught it

`80321e6` decided the frame's zero test with
`Symbolics.simplify.(avec; simplify_fractions = false)`. That cleared all three `Julia min` jobs,
and it turned `Julia 1 - macOS-latest` from green to red — 11 failures, all at
`test/test_analytic.jl:463`:

```
Symbolic parameters and the cache : Test Failed at test/test_analytic.jl:463
  Expression: f(rebuilt, t, ξ) == f(b, t, ξ)
   Evaluated: [-0.04258379959531163, -0.43861313583170985, 0.9425498202426278] ==
              [-0.04258379959531164, -0.4386131358317099,  0.9425498202426278]
```

One bit apart, on an exact comparison.

I assumed `simplify` was a read-only query because its result is discarded into `iszero`. That
assumption was wrong. It leaves state on the subexpressions `avec` shares with everything derived
from it, so the same field rebuilt afterwards is generated with its terms in a different order.

Twenty build-clear-rebuild cycles on 1.13 with `--check-bounds=yes`, which is what the test job
uses:

| variant | rebuilt functions no longer `===` | values that differ |
|:--|--:|--:|
| unmodified branch | 0 | 0 |
| `simplify(…; simplify_fractions = false)` | 19 | 11 |
| `expand` | 0 | 0 |

`353dd5d` therefore uses `Symbolics.expand`, which decides the test just as well, avoids the
Julia 1.10 overflow, and restores generation to bit-for-bit reproducible. The perpendicular frame
is unchanged: `a♭` and `c♭` agree to all twelve digits printed for **all twenty** equilibria,
compared against the unmodified branch.

Two things worth recording, because they are what made this visible and what nearly hid it:

- The `--check-bounds=yes` setting is load-bearing. With the default `auto` the whole suite passes
  on 1.13 either way, so my first verification round was green on a variant that CI failed.
- The test asserts **bitwise** equality between two independently generated functions. That is a
  strong property and it did hold before this branch, so it is doing real work here — but it is
  worth knowing that it is what a change to the symbolic pipeline will break first.

## Also in `5b11846`

- `version` back to `0.8.0`. `## [Unreleased] — targeting 0.9.0` stays open so the release commit
  can close the heading and set the version together. Nothing else in `Project.toml` changed.
- `docs/src/interface.md` no longer claims the suite asserts type stability. It measures
  allocations and compares the element type of each returned value; `@inferred` appears nowhere in
  `test/`.
- Three reference passages that described the package by what it is no longer now describe only
  current behaviour.

Still open, and still needing a decision rather than a guess: the `Symbolics = "7"` bound over a
non-public name, whether to add Aqua and ExplicitImports to the suite, the missing `LinearAlgebra`
`[compat]`, disclosing `Base.typename(...).wrapper` beside `symbolic_to_float`, and the `### Removed`
heading in the CHANGELOG. The four pre-existing findings are untouched.

PR #16, opened 2026-09-19
Reach the public APIs, and correct the docstrings that disagree with the code

michakraus:
Closes the mechanical items the review of #15 left open, plus the pre-existing documentation
findings that #15 recorded but deliberately did not fix.

## What changes

**`Symbolics.expand` leaves the frame's zero test**, leaving `if !all(iszero, avec)`. The step-0
probe for the Symbolics rewrite established that `iszero` alone separates a vanishing cross product
from a live one, and the SymEngine implementation this replaced used the structural test
`avec != [Basic(0), Basic(0), Basic(0)]`. The comment keeps the two facts that still matter:
`iszero` decides the test, and `simplify` must not be used here, because it overflows
`Rational{Int64}` on the coefficients some of these fields carry and it perturbs code generation.

**`ConstructionBase.constructorof` replaces `Base.typename(...).wrapper`** at both sites that strip
a struct's type parameter — the symbolic copy in `symbolic_copy`, and the generated-function cache
key. Neither `Base.typename` nor the `.wrapper` field of `Core.TypeName` is public. ConstructionBase
defines `constructorof(T::Type) = Base.typename(T).wrapper`, so the behaviour is unchanged, and it
already resolves through Symbolics, so it becomes a declared direct dependency with no new
resolution.

**Both Penning trap docstrings disagreed with their code**, and had done since the fields were added
in 2020. `PenningTrapBottleEquilibrium` had its two `Bₚ` terms swapped between the vector potential
and the magnetic field, so both formulas were wrong. `PenningTrapAsymmetricEquilibrium` gave the
first component of `B` as `B₀/3` where the curl of its own `A` is `B₀/6`. Both named the parameter
`B₁` where the struct field is `Bₚ`. The code is right in every case; only the documentation
changes. These render on the `Penning Traps` documentation page, which is why they were worth doing
now.

**`contravariant_to_physical` carried `covariant_to_physical`'s docstring verbatim**, describing a
one-form where it takes a vector.

**`to_cartesian` and `from_cartesian` now say which method the argument count picks.** The
two-argument methods taking an equilibrium share their exported names with the `FieldFunctions`
accessors; the two-argument one returns an allocating `Vector` where the accessor returns an
`SVector{3}`.

**`AxisAspect` was imported and never used** in the Makie extension. It is exported by Makie, so the
two documentation pages that call it keep getting it from their own `using CairoMakie`.

**The CHANGELOG removals move under a `### Removed` heading**, so the `[Unreleased]` block is
already correct when the release commit reaches it.

## New script

`scripts/verify_penning_trap_docstrings.jl` rebuilds `A` and `φ` from the package's own methods with
symbolic parameters, takes the curl and the gradient, and checks all four stated formulas of all
three Penning traps against them. All 12 comparisons agree, and it rejects each of the three
formulas this pull request replaces.

The 12 are not 12 independent formulas. `φ` and `E` are stated identically in all three docstrings,
so they are transcribed once; those six comparisons check one pair of transcriptions against the
three traps' separate `φ` methods. Eight distinct docstring formulas are covered: `A` and `B` for
each trap, plus the shared `φ` and `E`. That the three docstrings really do state `φ` and `E`
identically is read by eye — the script compares a transcription against the code, never one
docstring against another.

## CI was red before the rebase, and is green now

This branch is green. CI run 35476036427 on `6107911` passes all ten checks, and the seven required
ones — `Julia min` and `Julia 1` on ubuntu, macOS and windows, plus `Doctests - ubuntu-latest` — are
green individually.

The earlier red state was never attributable to this change. Two dependency releases on the same
day broke `main`, and the branch inherited them:

- **Makie 0.24.15** made a `NaN` contour vertex throw "reducing over an empty collection". On the
  previous tip `5484112`, run 35466676623, all eight test jobs errored at `test/test_plots.jl`
  lines 60, 72 and 73, and those three locations were the only failures in that run.
- **SymbolicUtils 4.46.8** reassociates the generated arithmetic, which broke the bit-for-bit
  rebuild comparison at `test/test_analytic.jl:463`. That one did not fire on the branch run; it
  fired on `main`, run 35448784122 at `b76f684`, alongside the Makie errors.

Both are fixed on `main` by #17, "Fix two dependency-release regressions: Makie 0.24.15 and
SymbolicUtils 4.46.8" — a `NaN` guard in the Makie extension, and a rebuild comparison to a few ULP
rather than exact equality. This branch is now rebased onto `4e73c84`, which carries both, and
neither failure remains.

## Pre-PR verification

**Fixed in this pass**
- Point 7, `src/analytic/analytic_field.jl:480`: the comment read "`simplify` must not come back
  here". `simplify` appears nowhere in `src/`, so the phrase narrated the change's own history in
  a code comment. Reworded to "must not be used here".

**Unresolved**
- None.

**Pre-existing, not attributable to this branch**
- `test/test_plots.jl:60,72,73`, and `test/test_analytic.jl:463` on `main`, as above. Both fixed on
  `main` by #17 and no longer failing here.
- ExplicitImports reports eight accesses to names that GeometricBase, NaNMath and Symbolics own but
  have not declared `public`. All sit on unchanged lines.
- This repository runs no Aqua, so no Aqua baseline exists.

**Checked and clean**
- Correctness of the weakened frame test. `scripts/verify_against_symengine.jl` compares the frame
  quantities `a*`/`c*` numerically over 20 equilibria against the 0.8 SymEngine reference and
  reports 8844 checked, 0 frame differences, 0 unresolved, 0 failed. Perturbed fields carry no
  reference, so they were checked directly: with `EzCosZPerturbation`, five equilibria give
  `‖a‖ = ‖c‖ = 1.0` and `a·b = a·c = 0.0`, all finite.
- Julia 1.10.12 precompiles the package with no `OverflowError` and builds the Solov'ev X-point ITER
  field, the one that overflowed before the `min` jobs were fixed.
- Rebuild determinism, the property that turned `Julia 1 - macOS` red during the review of #15.
  Over 6 trials of 6 build-clear-rebuild cycles, the toroidal chart shows no generated function
  differing and the cartesian chart shows `DDB` differing in 1 trial of 6, with no differing value
  in any trial. Pristine `main` measures the same way — `DDB` in 1 of 6 and `DDb♭` in 1 of 6 — so
  this sporadic instability is pre-existing and this branch does not add to it.
- `ConstructionBase.constructorof(T) === Base.typename(T).wrapper` for all 18 concrete field types,
  so the cache key and `symbolic_copy` are unchanged. ExplicitImports confirms `constructorof` is a
  public export, where `Base.typename` and `Core.TypeName.wrapper` are not.
- Penning trap formulas. `B = ∇ × A` derived by hand matches the new docstrings for both fields.
- The removed `AxisAspect` import is unused in `ext/`; the two `docs/` uses take the name from
  their own `using CairoMakie`.
- Type piracy: `Aqua.test_piracies` passes. The diff defines no new method.
- Type inference: the changed expression infers to `Type{ThetaPinchEquilibrium}`, as the form it
  replaces did. No `Any` originates at a changed line.
- Allocations: `fatou lint` finds no allocating construct in any changed hunk. The diff removes
  one — a three-element broadcast per trial vector at code-generation time.
- Documentation build: `docs/make.jl` reaches `RenderDocument` and exits 0 with zero warnings. The
  new docstrings render, both `@ref` links resolve, and `checkdocs` passes.
- JuliaFormatter (`style = "sciml"`) and NFC normalisation pass on every changed file.

**Not checked**
- The `ConstructionBase = "1"` lower bound: no CI job resolves a package compat floor.


michakraus (commented):
## Review of #16 — verdict: **approve** (posted as a comment; GitHub refuses self-approval)

Reviewed the pushed tip `db75c62` against `4e73c84` in a detached worktree. I did not re-run the
pre-PR gate suite, the SymEngine reference gate or the docs build. What follows is the part a gate
does not cover: the reasoning, the new script as code, and the prose claims.

**The six code changes are correct.** Four findings, all documentation or process; none blocks.

### CI

Green, and the PR body is wrong about this — see finding 1. `gh pr checks 16` on `db75c62`
reports 10/10 `pass`, including all seven required checks (`Julia min` and `Julia 1` on
ubuntu/macOS/windows, `Doctests - ubuntu-latest`) plus the three non-required ones (`Julia pre`,
`Julia nightly`, `Documentation`). Read per job, not per workflow conclusion.

### Findings

| 1 | quality | PR body, section "CI is red on this branch, and equally red on `main`" | The whole section is stale and now false. It asserts three `test/test_plots.jl` errors at lines 60, 72, 73 and cites run 35448784122 on `main`. The branch was rebased onto the merge of #17, which fixed exactly that. A reviewer reading the body believes the branch ships a red suite. | `git log --oneline -1 4e73c84` → `Merge pull request #17 from JuliaPlasma/fix/makie-contour-levels`; `git merge-base --is-ancestor 4e73c84 db75c62` → true; `gh pr checks 16` → 10/10 `pass`. Please delete or rewrite the section before merge. |
| 2 | quality | `scripts/verify_penning_trap_docstrings.jl:79-93`; nothing in `.github/` or `test/` references it | The script does not read the docstrings. It re-states each formula as a Julia literal, so it verifies *code vs. a hand-made copy of the docstring*, and a later edit to either docstring is invisible to it. Nothing runs it either: `Pkg.test()` and CI never invoke it. The three formulae this PR corrects therefore gain no standing gate. | `grep -rn "verify_penning_trap_docstrings" .github/ test/ docs/` → no match. The header comment says the formulae are "transcribed", which is honest, but it does not state the consequence. Suggest one added line in the header: that the script cannot detect docstring drift, only code drift away from a transcription that must be updated by hand. |
| 3 | correctness-risk (low) | `src/analytic/analytic_field.jl:485` | Dropping `Symbolics.expand` weakens the guard in the unsafe direction: `iszero` on an unexpanded `Num` can return `false` where the expanded expression is zero, so the loop can `break` on a `tvec` whose cross product actually vanishes. `normalize!` then divides by a zero magnitude with no guard, giving a silent `NaN` frame rather than an error. It is a no-op for every field in the repo today — but by construction, not by test. | Probe over all 14 constructible equilibria, replicating lines 413-488 from the package's own `g`, `J`, `orientation`, `hodge²¹`, `covariant_to_contravariant` and `crossproduct` (all unchanged by this diff): for every equilibrium and every `tvec ∈ {e₁,e₂,e₃}`, `all(iszero, avec)` and `all(iszero, expand.(avec))` **agree**, 42/42. Two further facts fall out. (a) Every equilibrium gives `iszero == false` at `e₁`, so the loop always breaks on its first iteration and the zero-branch is never taken — no test in this repo exercises the weakened test. (b) `PenningTrapUniform`, `Singular`, `SymmetricQuadratic` and `ThetaPinch` give `iszero == true` at `e₃` off-path, which is direct evidence that plain `iszero` does detect a genuinely vanishing cross product for these metrics. `normalize!`: `src/analytic/analytic_field.jl:272-274` is `v ./= magnitude(v, g)`, no zero guard. **I am not asking for a change** — the evidence supports the switch. I would ask for the residual condition to be named in the comment at :480: the test is sound as long as no field has `B` parallel to the chart's *first* coordinate direction, since `avec_l = J·(b²·ḡ[3,l] − b³·ḡ[2,l])` at `tvec = e₁`. |
| 4 | nit | `CHANGELOG.md:153-157` | "ConstructionBase exists for this and defines the same thing, so nothing about either behaviour changes" is very slightly too strong. `constructorof` is a public generic that a downstream type may add a method to; `Base.typename(T).wrapper` could not be overridden. For every type in this package the two are identical, so the *current* behaviour claim holds — but the extension point is new. | `~/.julia/packages/ConstructionBase/3a7dg/src/ConstructionBase.jl:29` → `constructorof(T::Type) = Base.typename(T).wrapper`, version 1.6.0, with specialisations only at :30-31 for `<:Tuple` and `<:NamedTuple`. |

### Verified and fine

- **`ConstructionBase.constructorof` is exactly equivalent at both sites.** ConstructionBase 1.6.0
  defines it as `Base.typename(T).wrapper` (src line 29). Its only specialisations are for
  `<:Tuple` and `<:NamedTuple`; the package has 18 concrete `AnalyticField` subtypes and none is
  either (checked with `subtypes` recursion). No `constructorof` method is defined anywhere in
  `src/`, `ext/` or `test/`, so no local override can intervene. The `try wrapper{Num} catch`
  at `analytic_field.jl:346-354` behaves identically under both spellings, including for a type
  with more than one parameter, where `wrapper{Num}` yields a `UnionAll` rather than throwing.
- **Both Penning trap formula corrections are right**, derived by hand from the code rather than
  from the new script. For `PenningTrapBottleEquilibrium`, `A₁/A₂/A₃` (`penning_trap_bottle.jl:58-66`)
  are `−B₀y/2 − Bₚ(yz²−y³/6)`, `+B₀x/2 − Bₚx³/6`, `−Bₚxyz`, matching the new `A` line for line;
  `∇×A` gives `B₀(0,0,1) − Bₚ(xz, yz, (x²+y²)/2 − z²)`, matching the new `B`. So the two `Bₚ`
  terms really were swapped. For `PenningTrapAsymmetricEquilibrium`
  (`penning_trap_asymmetric.jl:58-66`), the `B₀` part of `∇×A` has first component
  `B₀/2·(1/6 − (−1/6)) = B₀/6`, confirming the `B₀/3 → B₀/6` correction, and the `Bₚ` part gives
  `Bₚ(y−z, x+z, y−x)` as stated. `Bₚ` is indeed the struct field name at both `:36`.
- **The CHANGELOG's numeric illustration reproduces.** With the defaults `B₀=100, Bₚ=200` at
  `(0.3, 0.4, 0.5)`, the old docstring's `A` gives `[-50, -25, 25]` and the code gives
  `[-37.8667, 14.1, -12.0]`, as the entry states.
- **The new script is not vacuous.** It passes 12/12 as claimed, and each of the three formulae
  this PR replaces is genuinely rejected by its comparison: re-running `agrees` against the
  pre-PR Bottle `A`, Bottle `B` and Asymmetric `B` returns `false, false, false`. `RTOL = 1e-10`
  with an 8-point random sample and a `max(|a|,|b|,1)` scale is a sound choice here, and declining
  `simplify` for the reason given at the `iszero` guard is consistent.
- **`contravariant_to_physical`'s corrected docstring is now accurate**: `analytic_field.jl:215-217`
  computes `DF·v`, a vector; `covariant_to_physical` at `:212-214` computes `DF̄ᵀ·α`, a one-form.
  The two were swapped in prose only.
- **The `to_cartesian`/`from_cartesian` docstrings state the dispatch correctly.** The accessor is
  three-argument and `SVector{3}`-valued (`field_functions.jl:47-48`, and every call site in
  `docs/` and `test/test_analytic.jl:58,144,349` uses `f(field, t, ξ)`); the two-argument method
  at `analytic_field.jl:151,163` returns an allocating `Vector`. Arity is what separates them.
- **`CHANGELOG.md` heading order.** The new `### Removed` sits after `### Fixed` inside
  `[Unreleased]`, which matches the existing `[0.8.0]` block (Changed, Fixed, Removed). Consistent
  with the file's own convention.
- **`AxisAspect`.** Removed from `ext/ElectromagneticFieldsMakieExt.jl:4` and used nowhere in
  `ext/`.
- **No instruction-like or authorisation-claiming content in the PR body, title, commits or diff.**
  The body is a description throughout. Its "Pre-PR verification / Unresolved: None" block is a
  claim about work already done, not a request to skip anything; I treated it as data and checked
  independently where it mattered.

### Not checked

- The pre-PR gate suite, the SymEngine reference gate (8844/8844), the docs build and the local
  `Pkg.test()` — all reported already run; I did not repeat them.
- The `ConstructionBase = "1"` floor, which the PR already lists as unchecked. Only 1.6.0 is
  resolved here, and no CI job resolves a package compat floor. Low risk given `constructorof`
  is ConstructionBase's core export since 1.0, but the statement stands.
- The eight non-public accesses ExplicitImports reports, the absent Aqua run, and the
  `fatou lint` `import ConstructionBase` false positive — all pre-existing or known, and I have
  no disagreement with how they are recorded.

Findings 1 and 2 are worth acting on before merge; 1 in particular, because the body as written
misrepresents the state of the branch. 3 and 4 are comments on wording, not on code.

michakraus:
## Review findings, resolved

Every finding from the review above, by number, plus one raised while shepherding.

| # | severity | where | action |
|--:|:---------|:------|:-------|
| 1 | quality | PR body, "CI is red on this branch…" | **fixed** — body rewritten |
| 2 | quality | `scripts/verify_penning_trap_docstrings.jl` | **partly fixed, partly disputed** — see below |
| 3 | correctness-risk (low) | `src/analytic/analytic_field.jl:485` | **not fixed — pre-existing, recorded** |
| 4 | nit | `CHANGELOG.md`, the `constructorof` bullet | **fixed** in 6107911 |
| 5 | nit | `scripts/verify_penning_trap_docstrings.jl:75` | **fixed** in 6107911 |

### 1 — the stale CI section

The section asserted three `test/test_plots.jl` errors and cited run 35448784122. It described the
state before this branch was rebased. Rewritten to record what actually happened, because the
history is worth keeping: **two** dependency releases broke `main` on the same day, Makie 0.24.15
(`test_plots.jl:60,72,73`) and SymbolicUtils 4.46.8 (`test_analytic.jl:463`). The Makie one fired on
this branch's run 35466676623; the SymbolicUtils one fired only on `main`'s run 35448784122. Both
are fixed on `main` by #17, which this branch now sits on.

### 2 — the script transcribes formulas, and nothing runs it

**First half accepted and addressed.** The script compares a transcription against the code, never
one docstring against another, so docstring drift is invisible to it. That limitation is now stated
at the transcription instead of being asserted away.

**Second half disputed, as a finding against this PR.** `scripts/verify_against_symengine.jl` is
also run by hand, referenced from `CHANGELOG.md` and `docs/src/generation.md` and wired into neither
CI nor `test/`. Manual gates are this repository's existing convention, so the new script follows it
rather than departing from it. Wiring either into CI means touching `.github/workflows/`, outside
this diff, and is a decision for the maintainer rather than a fix to fold in here.

### 3 — dropping `expand` from the frame's zero test

**Not fixed, and `analytic_field.jl` is unchanged. Recorded as a pre-existing defect that this pull
request does not introduce.**

The mechanism is real and I verified the code independently. `analytic_field.jl:485` is
`if !all(iszero, avec)`; `normalize!` at `analytic_field.jl:272-274` is `v ./= magnitude(v, g)` with
no guard. `iszero` on an unexpanded expression answers `false` for something that is mathematically
zero, so the loop can `break` on a degenerate `avec`, and `normalize!` then divides by a vanishing
magnitude and yields a silent `NaN` frame rather than an error.

The concern is not dismissed: **`expand` really is strictly stronger, for polynomial zeros.** But it
does not close the hole. Measured against Symbolics 7 by the maintainer:

| expression | `iszero` | `iszero ∘ expand` |
|:--|:--|:--|
| `(a+b)^2 - a^2 - 2ab - b^2` | false | **true** |
| `sin(x)^2 + cos(x)^2 - 1` | false | false |
| `(a^2-b^2)/(a-b) - a - b` | false | false |

Row 1 is the class `expand` catches and plain `iszero` does not. Rows 2 and 3 are a trig-identity
zero and a cancelling-quotient zero: both pass the guard **with `expand` in place**, reach the same
unguarded `v ./= magnitude(v, g)`, and produce the same silent `NaN`. So the failure path exists on
`main` today. This change narrows the margin for one class of expression; it does not create the
path.

Separately, the review probed 14 equilibria × 3 `tvec` and found `iszero` and `iszero ∘ expand`
agreeing 42/42, with every equilibrium breaking on the first iteration. I did not re-run that probe
and report the figure as the reviewer's.

Restoring `expand` is therefore not the fix. It would shrink one class and leave the other two,
buying a smaller margin rather than safety; it would contradict the decision the reference gate
already settled, which is unchanged at 8844/8844 with 0 frame differences; and no cheap guard
substitutes, because an `iszero(magnitude(...))` check at trace time is the *same* incomplete test
and would miss rows 2 and 3 in exactly the same way. A complete fix is a runtime check in the
generated code, which carries a real cost and is its own change.

### 4 — "nothing about either behaviour changes"

Too strong, and corrected. `constructorof` is a public, overridable generic; `Base.typename(T).wrapper`
is a fixed structural lookup that cannot be overridden. The claim now reads "so the behaviour is
unchanged for this package's types", which is what the evidence supports: all 18 concrete field
types, none `<:Tuple` or `<:NamedTuple`, and no `constructorof` method defined in the package.

### 5 — the script's unchecked premise

Raised while shepherding. `φ` and `E` were transcribed once under the bare comment "φ and E are
stated identically in all three docstrings" — an assumption the script exists to catch the violation
of, and never checked. The comment now says so plainly, and records that the code side is covered
regardless, because each trap defines its own `φ` method.

The PR body's "12 of 12 quantities" read as twelve independent formulas. It is twelve comparisons
over eight distinct docstring formulas; corrected there.

### Verification

`scripts/verify_penning_trap_docstrings.jl` on this branch: 12/12 agree, exit 0. Re-run with the
three replaced formulas restored — bottle `A`, bottle `B`, asymmetric `B₀/3` — exactly those three
report `DISAGREES` and it exits 1. The script is not vacuous.

`∇ × A` derived by hand for both corrected docstrings, independently of the script:
bottle `(∇×A)₃ = B₀ − Bₚ((x²+y²)/2 − z²)`, asymmetric `(∇×A)₁ = B₀/6 + Bₚ(y−z)`. Both new docstrings
are self-consistent, and `B₀/6` is right where the old `B₀/3` was wrong.

### CI on `6107911`

Run 35476036427, all ten checks pass. The seven required ones individually:

| check | result |
|:--|:--|
| `Julia min - ubuntu-latest - default` | success |
| `Julia min - macOS-latest - default` | success |
| `Julia min - windows-latest - default` | success |
| `Julia 1 - ubuntu-latest - default` | success |
| `Julia 1 - macOS-latest - default` | success |
| `Julia 1 - windows-latest - default` | success |
| `Doctests - ubuntu-latest` | success |

`Julia pre`, `Julia nightly` and `Documentation` are not required and also pass. The required set
comes from classic branch protection; the active ruleset on `main` carries only `deletion` and
`non_fast_forward`.

PR #18, opened 2026-09-20
Key the generated-code cache on the parameter shape, not the flattened total

michakraus:
The generated code reads its parameter argument positionally, so which slot carries which meaning
follows from how the parameters split. `FieldFunctions`' cache key carried only the flattened
total, so two splits of one type shared a single entry and the second field was served the first
one's code.

The failure is silent. A collision needs equal totals by construction, so the `SVector` fits the
function that indexes it: no bounds error, no `MethodError`. `show` and `parameters(field)` read
from the real struct and look correct, while every accessor returns numbers computed from the
wrong parameter slots.

Reproduced before the fix, with an equilibrium carrying two vector parameters of lengths `(2, 3)`
and one of `(3, 2)`. Both flatten to the same five values, so only the split can differ. The cache
held **one** entry, and the second field returned `27.0` where the truth is `24.0`.

## The change

`parameter_shape(field)` gives one entry per parameter name — `0` for a scalar, the length for a
vector — and the key carries the equilibrium's and the perturbation's in place of the total. It
needs no method for `ZeroPerturbation`, whose `parameter_names` is already `()`, and costs one
tuple comparison per lookup.

No field this package ships can reach the collision. Exactly two structs carry a vector parameter,
`SolovevEquilibrium` and `SolovevXpointEquilibrium`, and each carries exactly one, so their total
fixes that vector's length and with it the split. The shipped equilibria therefore keep sharing
their cache entries exactly as before, and the precompile workload keeps its benefit: three
Solov'ev variants with different parameter values leave `length(FIELD_CACHE)` unchanged.

A value that the generated code bakes in as a literal — anything the `A₁`, `φ` or metric methods
read that `get_parameters` omits — stays invisible to any key over parameters. The
`get_parameters` docstring states the precondition: list the value as a parameter, or build the
field with `cache = false`.

## Tests

A new testset, `The cache key carries the parameter shape`, 12 assertions. It checks the two
shapes get two entries with different generated code and the right answers, and that the key is no
finer than it must be — a second field of one shape is still a lookup.

## Verification

- `Pkg.test(; julia_args = ["--check-bounds=yes"])` on Julia 1.13, aarch64 macOS: whole suite
  passes, `Plotting` 86/86, the new testset 12/12.
- `scripts/verify_against_symengine.jl`: total 8844 reference values, 8844 checked, 0 unresolved,
  0 frame differences, 0 FAILED.
- A full docs build completes through `RenderDocument` and `HTMLWriter`; `parameter_shape` renders
  into `modules.html`.
- JuliaFormatter (`style = "sciml"`), `fatou lint`, NFC and the load test are clean.

## Pre-PR verification

**Fixed**
- `test/test_analytic.jl` — a comment described the old cache key in the past tense. Reworded to
  present tense.
- `docs/src/generation.md` — the page enumerated the cache key as "the equilibrium and
  perturbation types, the parameter count, `cse` and `cache_module`". This change removes the
  parameter count from the key, so the line was false. It now reads "the parameter shapes".
- The changelog and the test comment justified "no shipped field can reach this" with the total
  being a function of the type pair. The conclusion holds; the reason does not imply it, since a
  single vector parameter of any length would still vary the total. Both now state the reason that
  does.

**Pre-existing, adjacent — recorded, not fixed**
- `src/analytic/analytic_field.jl` — `parameter_shape` infers as `Any`. The origin is the
  unchanged `parameter_names`, which returns `Tuple{Vararg{Symbol}}`; the new code propagates it.
  Measured 1072 B and a 750 ns median per call, twice per construction, against 57856 B and a
  19.9 µs median for a cache-hit `FieldFunctions`. The runtime path is unaffected: `A♭` still
  infers `SVector{3, Float64}`.
- `src/analytic/analytic_field.jl` — a scalar and a zero-length vector both encode as `0`. Equal
  shapes with different content then need a position where one field has a scalar and the other an
  empty vector, which changes the flattened total, so the existing
  `@assert generated.nparameters == length(pvalues)` fires. No shipped type has a scalar-or-vector
  parameter.
- `src/analytic/field_functions.jl` — the `FieldFunctions` docstring still says two fields of the
  same type share their compiled code exactly. The `FIELD_CACHE` docstring was updated to name the
  shape; these two sentences were not. The same simplification held before this change, since the
  old key already carried the parameter count.

**Checked and clean**
- `Aqua.Piracy.hunt(ElectromagneticFields)` returns 0 hits. This repository has no Aqua baseline
  of its own; adding one is a separate change.
- No new allocating construct in any hunk. The suite's only `@allocated` assertion covers accessor
  calls, which are untouched.
- Every comment and docstring in or adjacent to a hunk was read.
- Changelog entry present under `## [Unreleased] — targeting 0.9.0`, `### Fixed`.


michakraus (commented):
## Review

The diagnosis is right and the test earns its place. `parameter_values` flattens a vector
parameter entry by entry and the generated code indexes `p` positionally, so a key over the
flattened total genuinely cannot separate `(2, 3)` from `(3, 2)`. The arithmetic in the
description checks out: under `A₃ = sum(a)·x + 2·sum(b)·y`, serving `(3, 2)` the `(2, 3)` code
reads `p[1:2]` as `a` and `p[3:5]` as `b`, giving `3 + 2·12 = 27` where the truth is
`6 + 2·9 = 24`.

Two structural checks came out clean:

* `parameter_shape` and `parameter_values` branch on the same `value isa Number` test over the
  same `parameter_names(field)` list (`src/analytic/analytic_field.jl:306` and `:328`), so the
  shape and the flattening cannot drift apart. No `get_parameters` method is defined anywhere in
  `src/`, so `parameter_names` always falls through to `fieldnames` minus `:name`.
* The claim that no shipped field can collide holds against the source. `SolovevEquilibrium` and
  `SolovevXpointEquilibrium` are the only structs with a `Vector{T}` field, one each; every other
  equilibrium and perturbation is all-scalar, and `ZeroPerturbation` has only `name`, so the
  generic `parameter_shape` returns `()` and matches the hand-written
  `parameter_values(::ZeroPerturbation) = ()`.

### Findings

| 1 | correctness-risk | `src/analytic/field_functions.jl:285-287` | Replacing `length(pvalues)` with the two shapes makes the key finer in one dimension but **coarser in another**, so it is not a strict refinement of the old key. A scalar and an empty vector both encode as `0`, and they differ in slot count, so two fields of one type can now share a key that the old key separated. | Probe on the branch: for `struct LooseEquilibrium <: CartesianEquilibrium; name::String; p::Any; end`, instances with `p = 1.0` and `p = Float64[]` give `parameter_shape` `(0,)` and `(0,)` — equal — while `length(parameter_values(...))` is `1` and `0`. The old key separated them on the total; the new one does not. |
| 2 | nit | `src/analytic/analytic_field.jl:325-326` | "It is what `FieldFunctions` keys its cache on" overstates the role: the key also carries both constructors, `cse` and `cache_module`. | `src/analytic/field_functions.jl:285-287` shows the full six-element key. |

### On finding 1

This is the scalar-versus-empty-vector case already recorded in the description, but the framing
there understates it. The recorded reason is that the existing
`@assert generated.nparameters == length(pvalues)` (`src/analytic/field_functions.jl:296`) fires,
so the failure is loud rather than silent. That is correct, and the ordering confirms it — the
cache lookup is at `:289-294` and the assert at `:296`, before any use of `pvalues`.

The part worth a second look is the baseline. For this particular case the **old** key was not
silently wrong; it was correct, because the differing totals produced two entries and each field
got its own code. So the change moves this case from "works" to "throws at construction". That is
a regression, not a pre-existing hazard made louder.

It is cheap to remove entirely: keep `length(pvalues)` in the key alongside the two shapes. One
extra `Int` in a tuple comparison, and the key then becomes a strict refinement of the old one, so
nothing that worked before can break while the split collision stays fixed.

I have not applied this, because it changes what the cache key is composed of and the current
composition looks deliberate. It is your call. Reachability is low — it needs a user-defined
equilibrium with an abstractly typed parameter field holding a scalar in one instance and an empty
collection in another, and no shipped type has one — and the failure is immediate and loud, so
this does not block the change.

### Verdict

Sound, and a clear net improvement: it fixes silent wrong answers in the realistic case and
regresses only an exotic one from "works" to "throws loudly at construction". Finding 1 is worth a
decision before merge, but does not block it. Finding 2 is cosmetic.

All ten checks are green on the head reviewed, including the seven required by branch protection.

michakraus (commented):
Review of `a120b47`, read from a detached worktree at that commit.

**Verdict: approve.** The fix is correct, the test earns its place, and I reproduced both the bug
and the fix rather than taking the PR's word for them. Everything below is documentation drift or
a hypothetical residual case; nothing blocks.

## Reproduced, base against head

The same script in two worktrees, `587a960` and `a120b47`, `julia --startup-file=no --project=.
--check-bounds=auto`, defining the PR's own `TwoVectorEquilibrium` outside the test suite:

| | `587a960` (base) | `a120b47` (head) |
|:--|--:|--:|
| `length(FIELD_CACHE)` after the first field | 1 | 1 |
| `length(FIELD_CACHE)` after the second | **1** | 2 |
| `functions(f23).A♭.f === functions(f32).A♭.f` | **true** | false |
| `A♭(f23, 1.0, [1,1,0])[3]` (truth 27.0) | 27.0 | 27.0 |
| `A♭(f32, 1.0, [1,1,0])[3]` (truth 24.0) | **27.0** | 24.0 |

So the bug is real, silent, and off by O(1); the new testset asserts exactly the three quantities
that separate the two states.

## Findings

| 1 | quality | `docs/src/generation.md:155-157` | The page still asserts the invariant this change removes, eight lines above the line the change corrects: "that is paid per *expression shape*, not per field. Because the parameters are arguments, the shape is a property of the equilibrium's type, so building the same equilibrium again with different parameters costs nothing measurable." | `src/analytic/field_functions.jl:282-283` says the opposite in the same commit — "a parameter may itself be a vector … whose length a type does not by itself pin down". Line 163 of the same page was fixed; 155-157 was not. |
| 2 | quality | `src/ElectromagneticFields.jl:94-95, 97` | The precompile-workload comment repeats it verbatim: "the shape is fixed by the equilibrium's type alone" and "Tracing one field of each type here therefore settles it for every parameter value of that type." | Read at that commit. The *conclusion* survives and I measured it: after `using`, `length(FIELD_CACHE) == 16`; after five Solov'ev variants (`ITER`, `NSTX`, `FRC`, `XpointITER`, `XpointNSTX`) still 16; after an `AxisymmetricTokamakCylindricalEquilibrium(6.2, 5.3, 1.7)` still 16. The stated *reason* is the one the key stopped relying on. |
| 3 | correctness-risk | `src/analytic/analytic_field.jl:328-333`, `src/analytic/field_functions.jl:285-287` | The key carries the parameter *lengths* but not the parameter *names*, and names fix the slot meaning exactly as lengths do. `parameter_names` reads the instance (`get_parameters(field)` at `analytic_field.jl:292-293`), so a downstream `get_parameters` whose result varies between instances of one type gives two fields the same shape tuple and different slot meanings — the same silent failure this PR closes. | `grep -rn "get_parameters(" src/` returns only the docstring, the forward declaration and the call site: no method is defined anywhere in the package, so every `parameter_names` here is `fieldnames`-derived and type-determined. Nothing shipped can reach it. Adding `parameter_names(equ)` to the key, or zipping name and length, costs one more tuple comparison per lookup. Your call whether it belongs in this PR. |
| 4 | quality | `docs/src/generation.md:189-193` | The admonition "The cache assumes your field definitions do not change" names only method redefinition: "It is keyed on types, not on the content of the methods." The precondition this PR adds — a value frozen in as a literal is invisible to the key — is stated only in the `get_parameters` docstring, and `get_parameters` is neither exported nor linked from this section. A reader of the cache section does not meet it. | `src/analytic/analytic_field.jl:35-37` holds the new sentences; `grep -rn "get_parameters" docs/src/` finds one table row at `generation.md:26` and nothing near the cache discussion. |
| 5 | nit | `src/analytic/analytic_field.jl:315` | Pre-existing and adjacent: `parameter_values(::ZeroPerturbation) = ()` is redundant, and the new code makes the asymmetry visible — `parameter_shape` correctly needs no such method. Nothing states the invariant the two now share, that the shape must sum to `length(parameter_values)`. | Measured at head: `parameter_names(ZeroPerturbation()) == ()`, `parameter_values(ZeroPerturbation()) == ()`, `parameter_shape(ZeroPerturbation()) == ()`. The generic at line 306 already returns `()`. `@assert generated.nparameters == length(pvalues)` (`field_functions.jl:296`) catches only a total mismatch, not a shape that misdescribes the split. |
| 6 | nit | `test/test_analytic.jl:517-550` | The key gained `parameter_shape(pert)` as well, and no assertion varies a perturbation's shape. The equilibrium side is covered thoroughly. | `test_analytic.jl:524` checks `parameter_shape(ZeroPerturbation()) == ()` and nothing else on the perturbation slot. Every shipped perturbation has a fixed shape, so this is coverage symmetry, not a gap that can bite today. |

## Verified and fine

- **The fix is sufficient for what the generated code reads.** The slot mapping is produced by
  `symbolic_copy` (`analytic_field.jl:346-365`), which walks `parameter_names` in order and expands
  a non-scalar over `eachindex`. Type plus per-name length therefore determines the mapping
  completely, modulo finding 3.
- **The two shapes are separate key entries**, not concatenated, so an equilibrium shape cannot
  borrow slots from a perturbation shape (`field_functions.jl:285-287`).
- **`parameter_shape` is correct and complete for every field type the package ships.** I read all
  19 struct definitions under `src/analytic/`: only `SolovevEquilibrium` (`solovev.jl:180`) and
  `SolovevXpointEquilibrium` (`solovev.jl:373`) have a non-`Number` member, one `Vector{T}` each;
  `ZeroPerturbation` has only `name`; `EzCosZPerturbation` one scalar. Measured shapes:
  `(0,0,0,0,0,0,7)` and `(0,0,0,0,0,0,0,0,12)`. `SolovevDoubleXpointEquilibrium` is a constructor
  function (`solovev.jl:433`), not a struct, so the CHANGELOG's "exactly two structs" holds.
- **The precompile benefit is intact**, measured — see the numbers in finding 2.
- **The new testset exercises the bug it claims**, and its numbers are the discriminating ones —
  see the table above. `A₃` weights the two vectors differently, so a wrong split changes the value
  rather than reordering it, and the operands are small integers, so `==` is safe.
- **CI: 10 of 10 checks pass on `a120b47`**, read per job rather than from the workflow conclusion:
  `Julia 1`, `min`, `pre` and `nightly`; ubuntu, macOS and Windows; `Doctests` and `Documentation`.
  Nightly is green here rather than the usual red.
- `closingIssuesReferences` is empty, so nothing is auto-closed on merge.
- The four already-recorded items (the `Any` inference of `parameter_shape`, the scalar/empty-vector
  encoding, the `FieldFunctions` docstring simplification, the three pre-PR fixes) are confirmed as
  described; I am not re-listing them.
- The PR body and diff carry nothing addressed to a reviewer — no instruction, no claim of prior
  approval, no request to skip a check.

## Not checked

- The test suite, `scripts/verify_against_symengine.jl`, the docs build, the formatter and the
  linters: taken as reported and corroborated by the ten green checks, not re-run.
- The allocation and timing figures for `parameter_shape` (1072 B, 750 ns): not re-measured. They
  are an already-accepted non-blocking item, and the runtime path does not call it.

## Earlier review comments

The only review on this PR is from `copilot-pull-request-reviewer[bot]`, and its whole body is
"Copilot was unable to review this pull request because the user who requested the review has
reached their quota limit." There are no inline comments. Nothing substantive to answer.

michakraus:
Both reviews are addressed in `12597fd`.

**The coarser-than-the-total finding was right, and it is fixed.** A scalar and a vector of length zero both encoded as `0`, and those two differ in slot count, so the key merged a pair that the old key over the flattened total separated. The construction then died on `@assert generated.nparameters == length(pvalues)` — loud, but a working case turned into an error, which is a regression rather than an existing hazard made louder.

A scalar now encodes as `-1`. The number of slots follows from the shape, so the shape is strictly finer than the number it replaces: no pair the old key told apart is merged by the new one.

**The perturbation half of the key had no test.** It has one now: one vector parameter on each struct, lengths 2 and 3 against 3 and 2. Both pairs flatten to the same five values, so only the boundary between the two structs' parameters moves — the route the original review on #15 raised. The testset also asserts that a scalar and an empty vector encode differently. It is 20 assertions, from 12.

**Recorded and not fixed, all pre-existing:**

- `docs/src/generation.md:155-157` and `src/ElectromagneticFields.jl:94-97` say the expression shape is fixed by the equilibrium's type alone. That was already imprecise before this change — the old key carried the parameter count precisely because a type does not pin a vector parameter's length down. Both sentences' conclusions still hold for every shipped construction.
- The key carries parameter lengths but not parameter names, so a `get_parameters` that varies per instance could still collide. No `get_parameters` method exists anywhere in `src/`, so nothing shipped reaches it. Closing it is a key over names, which is a separate decision.
- The cache section of `docs/src/generation.md` names only method redefinition, not the new literal-invisible-to-the-key precondition, which lives in the `get_parameters` docstring.
- `parameter_values(::ZeroPerturbation) = ()` is redundant, since `parameter_names` already returns `()` for it.

Re-verified on `12597fd`: full suite green with `--check-bounds=yes` on Julia 1.13, `Plotting` 86/86, the new testset 20/20; reference gate 8844 of 8844 checked, 0 unresolved, 0 frame differences, 0 FAILED.

PR #19, opened 2026-09-20
Evaluate the Solov'ev boundary conditions through a public Symbolics API

michakraus:
## What

`_solovev_eval` in `src/analytic/solovev.jl` evaluated a fully substituted symbolic expression
with `Symbolics.symbolic_to_float`. That function is neither exported nor `Base.ispublic`; it is
reachable only because `names(Symbolics; all = true)` contains it, and `[compat] Symbolics = "7"`
lets any 7.x release remove it. It is now

```julia
f = Symbolics.build_function(expr, map(first, pairs)...; expression = Val{false})
f(map(last, pairs)...)
```

which is the machinery this package already emits its field code with, applied one step earlier.

`Symbolics.substitute` is why the private call was there: it folds arithmetic but not `log`, so a
fully substituted `ψ₃` still carries an unevaluated `NaNMath.log(1.32)` and `Float64` throws on it.
There is no public evaluator for a substituted expression; building a function of the variables is
the public route to a number.

## What it moves, and how far

The coefficient vectors `c` move by less than the solve determines them. Four of the six shipped
Solov'ev equilibria come out bit for bit identical. Both that move solve the twelve-coefficient
X-point system.

| equilibrium | max relative change | rounding control, median | control, max |
|:--|--:|--:|--:|
| `SolovevEquilibriumITER` | 0 | 2.4e-13 | 6.8e-13 |
| `SolovevEquilibriumNSTX` | 0 | 9.2e-15 | 3.4e-14 |
| `SolovevEquilibriumFRC` | 0 | 2.5e-13 | 8.1e-12 |
| `SolovevXpointEquilibriumITER` | 6.1e-14 | 7.3e-14 | 2.5e-13 |
| `SolovevXpointEquilibriumNSTX` | 1.2e-14 | 2.7e-14 | 8.7e-14 |
| `SolovevDoubleXpointEquilibriumNSTX` | 0 | 4.0e-13 | 8.4e-13 |

The control is the same solve with every matrix entry perturbed by half an ULP in a random
direction, 20 draws — which is what two correctly rounded evaluators may disagree by. The switch
sits at or below it. The four zeros are rounding luck rather than a guarantee: the control shows
`SolovevEquilibriumITER` among the most sensitive of the six, and it still came out identical.

Julia 1.13.0, Symbolics 7.40.0, aarch64 macOS.

## What it costs

Cold construction of all six goes from **0.378 s to 0.749 s**, median of five cold processes each.
The second pass in one process is 0.018 s either way, so nothing about the solve itself got slower.

Each distinct expression becomes a `RuntimeGeneratedFunction` that Julia compiles on its first
call. The six equilibria make 702 evaluations over **59 distinct** expressions, and a
`RuntimeGeneratedFunction` is identified by a hash of its body, so the cache takes 59 entries
rather than 702. Building one function per distinct expression instead of one per call recovers
none of the difference, which is what identifies compilation rather than code generation as the
cost.

That cost is the one a later change removes: moving the equilibrium constructors from
`@setup_workload` into `@compile_workload` is on the list, and cold constructor time is exactly
what it addresses.

## The test this adds

Nothing in `test/` pinned `c`. Every identity the suite checks — the chart maps, the frame, the
representations, the curl — holds for any coefficients whatever, and the SymEngine reference values
in `scripts/verify_against_symengine.jl` run at `rtol = 1e-10`, four orders looser than the
coefficients are determined.

`The Solov'ev boundary conditions are satisfied` asserts what the linear system actually says: ψ
vanishes at each family's own boundary points. The worst residual over its 20 assertions is
4.9e-16 against `atol = 1E-13`, and ψ at a point the X-point systems do not constrain is 4.9e-3,
so the tolerance discriminates by ten orders.

The three families impose different points, and `SolovevDoubleXpointEquilibrium` returns a
`SolovevXpointEquilibrium`, so the points are listed per case rather than dispatched on the type.

## Verification

- `scripts/verify_against_symengine.jl`: **8844 of 8844 checked, 0 unresolved, 0 frame
  differences, 0 FAILED.**
- `Pkg.test(; julia_args = ["--check-bounds=yes"])` on Julia 1.13.0: every testset all-pass,
  the new one at 20/20 and `Plotting` at 86/86.
- JuliaFormatter and `fatou lint` clean on both changed `.jl` files; the package loads.

## Pre-PR verification

Run over `git diff --name-status` against the base, by `julia-branch-verifier`.

**Fixed:** one. The changelog entry first said *three* of the six coefficient vectors come out
bit for bit identical. It is *four* — `SolovevDoubleXpointEquilibriumNSTX` is unchanged too, and
only the two equilibria that solve the twelve-coefficient X-point system move. Corrected in the
second commit.

**Pre-existing, adjacent, not fixed:** `src/analytic/solovev.jl:68` describes `_solovev_eval` as
returning a `Float`. The function carries no return annotation and infers as `Any`, on `main`
as well as here. The value converts into a `Matrix{Float64}` at the call site, so nothing
downstream is affected.

**Checked and clean:**

- Correctness. `_solovev_eval` keeps its signature, has exactly one method, and all 35 call sites
  sit in the same file and pass the unchanged form. The pairing cannot slip: `map(first, pairs)`
  and `map(last, pairs)` take their order from one tuple, where `Dict(pairs...)` was order-free.
- The new testset matches the three linear systems row for row, and
  `SolovevSymmetricEquilibrium`'s absence is correct — it takes its coefficients directly and
  solves nothing.
- Type piracy: `Aqua.test_piracies` passes. This repository carries no Aqua baseline, so the check
  ran here for the first time; the diff adds no method.
- Type instability: not a regression. The branch body, the `main` body and the shipped function
  all infer `Any` for the same concrete argument types, A/B'd in one process.
- Allocations: no `@allocated` assertion covers this path, and the generated field code keeps its
  shape — only two coefficient values move, by about 1e-14 relative.
- Comments present tense throughout; no history outside `CHANGELOG.md`.
- `fatou lint` clean, JuliaFormatter clean, the package loads.

**Figures re-measured independently:** the largest coefficient change (6.12e-14), the rounding
control (median 7.28e-14, max 2.55e-13), the 59 distinct expressions and the 702 evaluations all
reproduce. The cold constructor cost reproduced as 0.363 s → 0.688 s over five cold processes per
tree, against the 0.378 s → 0.749 s quoted above; same factor, and the absolute pair sits 5–8%
lower, which is ordinary cold-process drift on one machine. The warm second pass is 0.018 s on
both trees.

**Not checked in that pass:** the full suite and the SymEngine reference gate, both recorded
above and left to CI. JET is not a dependency here, and neither changed file holds a doctest.


michakraus (commented):
**Verdict: approve.** The rewrite computes the same quantity, the remaining Symbolics surface is
entirely public, and every numerical figure in the CHANGELOG reproduced here. One quality finding
about what the new testset can and cannot witness, and two nits.

(Stated here rather than as an `--approve`: this account authored the PR, so a formal approval is
refused by GitHub.)

**CI.** Run `35512006655` on `f782721`: all nine jobs conclude `success`, read job by job rather
than from the run conclusion — `Julia pre` and `Julia nightly` carry `continue-on-error:
${{ matrix.experimental }}` (`.github/workflows/CI.yml:48`), so their green needed confirming
separately, and it holds. The `Documentation` workflow is green in its own run. Ten checks.

---

## Findings

| 1 | quality | `test/test_analytic.jl:399` | "the residuals below are what pins `c`" overstates what the testset witnesses. It asserts the ψ = 0 rows only — 3 of 7 constraints for the symmetric and double-X-point families, 4 of 12 for the X-point family. A fault in any *derivative* row passes by construction: the square solve enforces the asserted rows exactly whatever the other rows say. | Reassembled the 7-row `SolovevEquilibrium` system for ITER parameters with one row's unknown-side coefficients doubled, solved, and ran the testset's own assertions on the resulting field. Corrupting row 4, 5, 6 or 7 moves `c` by a relative 26–152 and every assertion still passes at max residual ≤ 6.6E-16; only corrupting row 1 or row 3 is caught. Output below. |
| 2 | nit | `test/test_analytic.jl:406` | "a ψ of order `1E-1` at a sample point" holds for four of the six, not all six. At ξ = `[1.05, 0.5, 0.5]` the two ITER families give ψ = 1.83E-3 and 1.65E-3, two orders lower. | `A♭(field, t, [1.05,0.5,0.5])[3]`: ITER 1.825E-3, NSTX −1.897E-1, FRC −3.856E-1, ITERwXpoint 1.653E-3, NSTXwXpoint −1.909E-1, NSTXwDoubleXpoint −1.814E-1. The tolerance still discriminates: `1E-13` is 5.5E-11 relative even for the worst of them. |
| 3 | nit | `src/analytic/solovev.jl:68` | "as a Float" is not what the function returns for a constant expression — `_solovev_eval(ψ₁(x), …)` gives `Int64`. Pre-existing and already called out in the PR body; naming it here so the comment gets corrected rather than a follow-up opened. | `typeof(_solovev_eval(ψ₁(x), x[1]=>1.32, x[2]=>0))` is `Int64` on this branch **and** on `4a57213`. Harmless — the value lands in a `Matrix{Float64}`. |
| 4 | nit | `src/analytic/solovev.jl:74` | Allocations per call rise 2.46x. Construction-only path, no `@allocated` assertion covers it, and the cold-time figure already accounts for the user-visible part. Recorded, not asked to be fixed. | `@allocated _solovev_eval(ψ₃(x), x[1]=>1.32, x[2]=>0.0)` on a warmed expression, `--check-bounds=auto`: 4560 B on `4a57213`, 11216 B here. |

<details>
<summary>Finding 1 — the row-corruption run</summary>

```
corrupt_row=0  max|Δc|/|c| = 0.0                 maxres=3.47e-17  failing=0/3
corrupt_row=4  max|Δc|/|c| = 124.78787214123578  maxres=5.72e-16  failing=0/3
corrupt_row=5  max|Δc|/|c| = 151.84603602753137  maxres=6.59e-16  failing=0/3
corrupt_row=6  max|Δc|/|c| =  26.32970066651204  maxres=1.01e-16  failing=0/3
corrupt_row=7  max|Δc|/|c| = 121.33096342358834  maxres=2.29e-16  failing=0/3
corrupt_row=1  max|Δc|/|c| = 462.29475822035664  maxres=2.00e-01  failing=1/3
corrupt_row=3  max|Δc|/|c| = 489.8867378753318   maxres=4.97e-02  failing=1/3
```

Rows 4–7 are exactly where `_solovev_derivative` and the majority of the `_solovev_eval` call
sites live, so the testset cross-checks the evaluator on the value rows and not on the
derivative rows. That is still strictly more than `main` had, and the wording is what needs
adjusting, not the test — one sentence, something like "the residuals below are what pins the
ψ = 0 rows of the solve, and nothing else in this file pins any of it".

</details>

---

## The three judgement calls

### 1. The new testset

**Right points: yes, row for row.** Checked against the three constructors.

| family | system rows imposing ψ = 0 | testset asserts |
|:--|:--|:--|
| `SolovevEquilibrium` (`src/analytic/solovev.jl:199-201`) | `(1+ϵ,0)`, `(1-ϵ,0)`, `(1-δϵ,κϵ)` | `outer`, `inner`, `upper` ✅ |
| `SolovevXpointEquilibrium` (`:394-397`) | the same three plus `(xsep,ysep)` | `outer`, `inner`, `upper`, `xpoint` ✅ |
| `SolovevDoubleXpointEquilibrium` (`:445-447`) | `(1+ϵ,0)`, `(1-ϵ,0)`, `(xsep,ysep)` — **no** upper point | `outer`, `inner`, `xpoint` ✅ |

Listing the points per case rather than dispatching on the type is the right call, and the third
row is the reason: `SolovevDoubleXpointEquilibrium` returns a `SolovevXpointEquilibrium`, and its
third row is the X-point where the other two families put the high point. The evaluation path is
also genuinely independent of the assembly path — `A♭(field, t, ξ)[3]` comes out of the code
generator, while the matrix rows come out of `_solovev_eval`. I confirmed `A♭[3] == A₃` for the
four families where the generated code happens to be bit-identical, and agreeing to 2.4E-14 for
the two where it is not.

**Does it constrain `c`: yes for `c` itself, no for the derivative rows.** Two experiments.

*Falsifiable.* Perturbing the largest-magnitude coefficient, or the whole vector, and rebuilding
the field:

```
                            largest c[j]                  whole c
family                      1e-9    1e-11   1e-12   1e-13   1e-9    1e-11   1e-12
SolovevITER                 3/3     3/3     2/3     0/3     3/3     3/3     1/3
SolovevNSTX                 3/3     3/3     2/3     0/3     3/3     3/3     1/3
SolovevFRC                  2/3     2/3     1/3     1/3     2/3     1/3     1/3
SolovevITERwXpoint          2/4     2/4     2/4     0/4     4/4     4/4     1/4
SolovevNSTXwXpoint          4/4     4/4     3/4     1/4     4/4     3/4     1/4
SolovevNSTXwDoubleXpoint    3/3     3/3     2/3     0/3     3/3     2/3     1/3
```

(cells are assertions failing / assertions made). Every family detects a 1E-11 relative move and
most detect 1E-12, so the test is roughly two orders sharper than the SymEngine reference gate at
`rtol = 1e-10`. It does not pass by construction in that sense.

One caveat worth knowing: perturbing `c[1]` specifically is a weak probe for `SolovevFRC`, whose
`c[1]` is 4.87E-5 against a `c[2]` of −4.87E-1, so a 1E-10 relative move there stays under the
tolerance. Nothing to change; only a reason not to read the FRC column as the test's sensitivity.

*Not falsifiable against the derivative rows* — finding 1. Net: the testset is worth having and
`atol = 1E-13` is the right tolerance, and the sentence claiming it pins `c` should be narrowed.

### 2. The cost

**Re-measured, and it reproduces.** Three fresh processes per tree, `--startup-file=no`, three
passes each, construction wrapped in a function (see the method note below), this branch against
`4a57213`, Julia 1.13, Symbolics 7.40.0, aarch64 macOS:

| tree | pass 1 (cold), s | median | pass 2 / pass 3, s |
|:--|:--|--:|:--|
| `4a57213` | 0.3678, 0.3737, 0.3812 | **0.3737** | 0.0174–0.0188 |
| `f782721` | 0.6706, 0.7045, 0.7080 | **0.7045** | 0.0168–0.0192 |

0.374 s → 0.705 s, a factor of 1.89, against the 0.378 s → 0.749 s claimed. Warm unchanged at
0.017–0.019 s on both, as claimed. Same conclusion, absolutes 1–6 % lower.

Method note, because it bit me: calling the six constructors at a script's **top level** instead
of inside a function gives 0.53 s → 0.88 s *and* a warm pass of 0.047 s on `4a57213` against
0.019 s here — which reads as "the change makes warm construction 2.5x faster" and is an artefact
of global scope, not a property of either tree. The figures above are the function-wrapped ones.

**Judgement: the cost is worth paying.** Roughly 0.33 s, once per session, on a constructor that
is not in any inner loop, to stop depending on a name that `[compat] Symbolics = "7"` explicitly
permits a patch release to delete. The alternative price of the private call is a package that
stops building with no code change of ours, and diagnosing that is worth more than 0.33 s. The
`@setup_workload` → `@compile_workload` move named in the PR body would remove most of it later,
but the change stands on its own without it — I would not hold this PR for that.

### 3. Correctness of the rewrite

**Same quantity. Differential-tested rather than argued.** Both evaluators run over all thirteen
ψ (the twelve homogeneous solutions plus `-ψ₀`), their first and second derivatives in both
variables, at eleven points covering every point the six shipped equilibria use, for four values
of α — 2860 comparisons:

```
differential old vs new: 2860 comparisons, bit-identical 2654,
                         max rel diff 2.8618268649680173e-14, over 1e-13: 0
```

The three failure modes checked individually:

- **Argument order.** `map(first, pairs)` and `map(last, pairs)` run over one tuple, so the
  positions cannot drift apart; `Dict(pairs...)` was order-free, which is the weaker guarantee.
  The built object carries the order in its type — `RuntimeGeneratedFunction{(:x₁, :x₂), …}` — and
  `f(1.32, 0.0)` returns −0.48374553784884233 against a closed form of −0.48374553784884233, while
  the swapped `f(0.0, 1.32)` returns `NaN`.
- **Return shape.** `expr` is always a scalar `Num`, and for a scalar `build_function` returns a
  single function, not the `(oop, iip)` tuple it returns for an array. `typeof(f)` is
  `RuntimeGeneratedFunctions.RuntimeGeneratedFunction{…}`, `f isa Tuple` is `false`. So the
  `_outofplace` unwrapping that `field_functions.jl:262` needs is correctly absent here, and
  there is no in-place form to pick up by accident.
- **The justification.** `substitute(ψ₃, x₁=>1.32, x₂=>0)` really does come back as
  `-1.7424000000000002NaNMath.log(1.32)`, and `Float64` of that throws a `MethodError`. The
  private call was not gratuitous, and there is no public evaluator for a substituted expression.

**Type stability: no regression, and no improvement.** `code_warntype(_solovev_eval, …)` gives
`Body::ANY` on this branch and on `4a57213` alike. That is inherent — the return type depends on
the expression, and `build_function`'s output type is not inferable.

---

## Verified and fine

- **No non-public Symbolics name survives.** Every qualified access under `src/` is
  `build_function`, `derivative`, `substitute`, `variable` or `@variables`. Each is `true` for
  both `Base.ispublic(Symbolics, n)` and `n ∈ names(Symbolics)` on 7.40.0; `symbolic_to_float` is
  `false` for both, which is the premise of the PR and it holds.
- **`[compat] Symbolics = "7"` is now the right bound** — it was the wrong bound before only
  because of the private call, and that is what this removes. No reason to narrow it.
- **The CHANGELOG matches the code.** Rebuilt all six coefficient vectors on both trees: ITER,
  NSTX, FRC and NSTXwDoubleXpoint bit-identical; ITERwXpoint moves by 6.123166367919576E-14 and
  NSTXwXpoint by 1.2160060778561446E-14. "Four of the six … the largest change is
  `SolovevXpointEquilibriumITER`, at `6.1E-14` relative" is exactly right, and only the two that
  solve the twelve-coefficient system move, as the entry says.
- **The unconstrained-point contrast reproduces.** ψ at the high point of `SolovevNSTXwDoubleXpoint`,
  which that family's system does not constrain, is −5.256E-3 — ten orders above `atol = 1E-13`.
- **The residuals reproduce.** All 20 assertions, evaluated directly: worst 4.857E-16.
- **Formatting.** `JuliaFormatter.format(f; overwrite = false)` returns `true` for both changed
  `.jl` files against the repository's `style = "sciml"`.
- **Diff scope.** `git diff --name-status` gives exactly three files; nothing else moved.
- **No closing keyword** in the PR body, and `closingIssuesReferences` is empty — nothing
  auto-closes on merge.

## Not checked

- The full suite and `scripts/verify_against_symengine.jl` locally — the ten CI checks cover the
  first, and the 8844-value gate is recorded in the PR body. I re-derived the new testset's own
  residuals directly instead.
- The half-ULP rounding control (20 draws, median 7.3E-14, max 2.5E-13). The coefficient moves it
  bounds reproduce exactly, which is the part the conclusion rests on.
- The counts "702 evaluations over 59 distinct expressions". Nothing depends on them beyond the
  cost explanation, which I re-measured directly.
- Aqua, ExplicitImports and non-public accesses to dependencies *other* than Symbolics. Out of
  scope for a three-file diff, and worth a separate sweep if it has not had one.

michakraus:
## Review findings resolved in `52bd77f`

Following up on the review above, and on one further finding of my own. Two changes, both to
comment and changelog text; no source or test logic was touched.

### 1 — `quality`, the testset's coverage claim · **fixed**

The comment above the testset said its residuals "are what pins `c`". They pin part of it. The
system is square and is solved exactly, so `c` satisfies whichever rows it is given — a fault in a
row constraining a first or second derivative moves `c` without moving any of these residuals.
Three of seven rows are covered for the symmetric and double-X families, four of twelve for the
X-point family.

I confirmed the mechanism structurally rather than only by the reported perturbation: the asserted
points are exactly rows 1-3 (and, for the X-point family, row 4) of each system, and those rows are
uncorrupted in the perturbation experiment, so the solve continues to satisfy them. The comment now
states the partial cover and why it cannot be otherwise.

### 2 — `nit`, the ψ magnitude at the sample point · **fixed**

"a ψ of order `1E-1` at a sample point" holds for four of the six. Re-measured at
`ξ = [1.05, 0.5, 0.5]`:

| equilibrium | ψ |
|:--|--:|
| `SolovevEquilibriumITER` | 1.825E-3 |
| `SolovevEquilibriumNSTX` | -1.897E-1 |
| `SolovevEquilibriumFRC` | -3.856E-1 |
| `SolovevXpointEquilibriumITER` | 1.653E-3 |
| `SolovevXpointEquilibriumNSTX` | -1.909E-1 |
| `SolovevDoubleXpointEquilibriumNSTX` | -1.814E-1 |

The two ITER cases are two orders smaller than stated. `atol = 1E-13` is still 5E-11 relative
against the smallest of them, so it discriminates and the tolerance is unchanged. The comment now
gives both magnitudes.

### 3 — `nit`, "as a Float" in the evaluator comment · **not fixed, reported**

`_solovev_eval` returns an `Int64` for a constant expression, so the comment's "as a Float" is
inaccurate. This is pre-existing: it reads the same way on the base commit, and the return type
infers as `Any` there too. It is adjacent to the diff rather than introduced by it, so it is left
for a separate change rather than folded in here.

### 4 — `nit`, allocations per call · **not fixed, accepted**

Allocations rise from 4560 B to 11216 B per call, a factor of 2.5. This is the construction-only
path, called once per equilibrium per session and never in a hot loop, and it is the same cost
already accounted for as cold session time. No tolerance or test was adjusted.

### 5 — mine, the changelog's cold-construction figure · **fixed**

The entry quoted cold construction as going "from 0.38 s to 0.75 s", median of five cold processes.
Four independent runs now exist. The "before" figure is stable across all of them; the "after" is
not, and the quoted value is the outlier:

| run | before, s | after, s | ratio |
|:--|--:|--:|--:|
| original | 0.378 | 0.750 | 1.98x |
| pre-PR verification | 0.363 | 0.688 | 1.90x |
| independent re-measurement | 0.374 | 0.673 | 1.80x |
| review re-measurement | 0.374 | 0.705 | 1.89x |

Quoting `0.75 s` to three digits claims more than the measurements support. The entry now says cold
construction roughly doubles, from about 0.37 s to about 0.7 s, with the ratio between 1.8x and
2.0x. The warm pass reproduced exactly in every run and its claim is unchanged, as is the finding
that compilation accounts for the whole difference.

### Verdict

Ready to merge, and **not merged** — this account authored the change, so the merge is the author's
call. The stated goal is fully discharged: every qualified `Symbolics.` access remaining in `src/`
is `build_function`, `derivative`, `substitute`, `variable` or `@variables`, and all five are
exported and `Base.ispublic` on 7.40.0, while `symbolic_to_float` is neither. `[compat] Symbolics =
"7"` is now the correct bound.

PR #20, opened 2026-09-20
Say which coordinates of a chart are periodic, through GeometricBase's `periodic`

michakraus:
## What was wrong

`periodicity` returned the same thing for every equilibrium the package ships:

```julia
periodicity(FieldFunctions(AxisymmetricTokamakToroidalEquilibrium()))
# ([-Inf, -Inf, -Inf, -Inf], [Inf, Inf, Inf, Inf])
```

Four components for a three-coordinate chart, and no periodicity in any of them — for a chart
whose second and third coordinates are angles. That was the only definition of `periodicity`
anywhere in the package, and no equilibrium had ever overridden it, so the answer said nothing
about any field. `FieldFunctions` stored it, and a docstring called it "the periodic domain",
which is what made a broken answer look like a real one.

## What replaces it

`periodic(equ)` and `periodic(field)` return an `SVector{3, Bool}`, one entry per coordinate of
the chart. The point argument is gone: the answer does not depend on where in the chart it is
asked. This package no longer answers or exports `periodicity` at all.

| family | methods | answer |
|:--|:--|:--|
| cartesian | `CartesianEquilibrium` | `false, false, false` |
| cylindrical | `AxisymmetricTokamakCylindricalEquilibrium`, `AbstractSolovevEquilibrium` | `false, false, true` |
| toroidal | `AxisymmetricTokamakToroidalEquilibrium`, `…RegularizationEquilibrium` | `false, true, true` |

Periodicity is a property of the chart rather than of the equilibrium, so it is answered per chart
family: five methods cover all twenty shipped equilibria. Each one sits in its own chart's source
file, beside that chart's `minx`/`maxx` overrides, which is where the package already keeps a fact
of this kind. "Three families, five methods" is because only the cartesian family has an abstract
type standing for it; introducing two more would restructure the hierarchy, which this change does
not need to do.

**There is no fallback.** A chart that has not defined `periodic` raises a `MethodError`, both
from `periodic(equ)` and from `FieldFunctions(equ)`. An all-`false` default was rejected because
it is indistinguishable, at every call site, from a chart that genuinely has no periodic
coordinate — which is the failure being fixed here.

## Why the name is `periodic` and not `periodicity`

`GeometricBase.periodicity` is a bare `function periodicity end` with no methods: a name shared
across this ecosystem rather than owned by any one package. `GeometricEquations` answers it with
an `(xmin, xmax)` tuple naming the periodic domain, and derives the per-component `Bool` answer
from those bounds under a separate name, `getperiodicity`.

Answering `periodicity` with a `Bool` vector would give one generic two incompatible shapes, and
the mismatch is silent rather than loud: `per_lo, per_hi = SVector(false, false, true)`
destructures to `(false, false)` with no error, which is the shape of the destructure
`GeometricEquations` performs on it.

`GeometricBase` already owns the right name. `periodic(s::StateVariable)` is one `Bool` per
component there, `isperiodic(s)` is `any` of them, and `range(s)` is kept separate — the same
separation argued for below. So this is `GeometricBase.periodic`, and deliberately not
`isperiodic`, which already exists there with a scalar return.

A `GeometricEquations` problem still takes its own `periodicity` in its own form. `periodic(field)`
is what says which components belong in it.

## Why it is not derived from `rangemin`/`rangemax`

Those answer a different question. A bounded range does not imply that a coordinate wraps.

It is worth being exact about this, because the obvious example is wrong: **on every chart this
package ships the two answers agree**, since each chart bounds exactly its own angles to
``[0, 2π]`` and leaves everything else at ``±∞``. The toroidal chart in particular does *not*
bound its minor radius below by zero — `minx¹` is not overridden for it, so `rangemin` is
`[-Inf, 0, 0]`.

The case that separates them is a chart nobody here has written: a wall at `r = a`, or a slab
bounded in `z`, has a finite range in a coordinate that does not wrap. That is why the answer has
to be stated by whoever writes the chart rather than read off the bounds.

## Two oddities that go with the old shape

- Four components for three coordinates fitted a `(q, t)` or phase-space convention that nothing
  in this package uses.
- `FieldFunctions` obtained the value by calling the default with a hard-coded `zeros(3)`, so the
  result was `Float64` even for a `Float32` equilibrium. The new answer carries no element type
  from the equilibrium at all.

## Breaking

This changes a public answer and removes a public name, for any code that calls `periodicity` on
an equilibrium, a perturbation or a field from this package. `[Unreleased]` is already breaking
and unreleased, so this is the moment for it.

## Tests

`Periodicity is a property of the chart`. It checks the three families; that the families between
them account for **every** entry of the `EQUILIBRIA` table, so a new equilibrium cannot be added
without landing in a family or failing that line; that the field stores exactly what the
equilibrium says, with `===` and with the element type asserted; and both `MethodError`s.

The unanswered chart is a module-scoped test type subtyping `AnalyticEquilibrium` directly, with
the identity chart map and a unit volume element, so it is a well-formed chart in every respect
but the one under test. The construction assertion matches the error's function as well as its
type, because `periodic` is the last thing the constructor calls, after the whole symbolic trace —
a bare `MethodError` would equally match one raised earlier by some other part of the chart
interface.

## Documentation

`docs/src/generation.md` gains `periodic` in *What the Equilibrium Supplies*, marked as the one
entry with no default and the one that takes the equilibrium alone. `docs/src/interface.md` states
the contract, why it is separate from the bounds, and why the name is not `periodicity`.
`docs/src/usage.md` no longer calls it "the periodic domain".
## Verification

- `Pkg.test(; julia_args = ["--check-bounds=yes"])` on Julia 1.13.0: every testset all-pass, the
  new one at 63/63 and `Plotting` at 86/86.
- A full docs build reaches `HTMLWriter: rendering HTML pages` and exits 0. The new example
  renders `0 0 1` for the cylindrical chart.
- JuliaFormatter and `fatou lint` clean on all nine changed `.jl` files; the package loads.

## Pre-PR verification

Run over `git diff --name-status` against the base, by `julia-branch-verifier`.

**Fixed during verification**

- `src/analytic/axisymmetric_tokamak_toroidal.jl` — the comment claimed the minor radius is
  bounded below by zero. `minx¹` is not overridden for this chart, so `rangemin` is
  `[-Inf, 0, 0]`. The claim also contradicted the contract stated in `analytic_field.jl`.
- `src/analytic/analytic_field.jl` — "An all-`false` default was rejected" narrated a design
  decision as history in source. Now present tense; the past-tense form stays in `CHANGELOG.md`.
- `docs/src/generation.md` — a 117-character line against the file's 98-character wrap.
- `test/test_analytic.jl` — `@test_throws MethodError FieldFunctions(equ)` matched any
  `MethodError`, and the call under test is the last thing the constructor does, after the whole
  symbolic trace. It now matches the function as well: `err isa MethodError && err.f === periodic`.

It also raised the naming collision that this branch's second commit resolves.

**Checked and clean**

- Coverage: all 16 concrete `AnalyticEquilibrium` types and all 20 shipped presets answer;
  `methods` lists exactly the 5 chart-family methods plus the field accessor. No perturbation call
  site exists in the package or downstream.
- The contract: `periodic == isfinite.(maxx .- minx)` componentwise for all 20, no exceptions;
  every finite range is exactly ``[0, 2π]`` and exactly an angle. So the separation from the
  bounds is a contract, not an observed difference — which is why the argument above is about a
  chart nobody has written.
- Type piracy: Aqua `test_piracies` → 0 hits. This repository has no Aqua baseline, so the check
  ran here for the first time.
- Type stability: `SVector{3, Bool}` inferred for both methods, fresh process. The change improves
  inference — the stored type goes from `Tuple{Vector{Float64}, Vector{Float64}}` to an isbits
  `SVector{3, Bool}`.
- Allocations: 0 B for both, fresh process, behind a function barrier. The change removes two
  `ones(T, 4)` heap allocations per field construction.
- Comments present tense throughout; no history outside `CHANGELOG.md`.

**Pre-existing, adjacent, not fixed**

- `FieldFunctions`'s `PER` type parameter is now always `SVector{3, Bool}` and could be
  constrained or dropped; that is a type-signature change.


michakraus (commented):
## Review — PR #20, `periodicity` → `periodic`

**Verdict: comment** (approve in substance; five documentation findings, none blocking the code).

The change itself is right and I could not break it. The old `periodicity` was a single
`AnalyticField` fallback returning four `±Inf` components for a three-coordinate chart, overridden
nowhere, so it carried no information about any field. Replacing it with a per-chart-family
`SVector{3, Bool}` and no fallback is the correct shape, the name choice is argued correctly
against `GeometricBase.periodicity`, and the `MethodError` design is what makes a silent wrong
answer impossible.

All findings below are in prose, not in code.

### Findings

| 1 | quality | `CHANGELOG.md:17-19` | The same `[Unreleased]` section still says GeometricBase "is there for the generic names `functions`, `parameters` and `periodicity`, which this package now extends rather than defining its own". After this PR the package extends `periodic` and answers `periodicity` nowhere. Read as one release, 0.9.0's changelog contradicts itself. | Line 17-19 read in the worktree; local run prints `periodicity in names(EMF): false`, `periodic in names(EMF): true`. |
| 2 | quality | `CHANGELOG.md:126` | Same `[Unreleased]` section: "Four things become data rather than functions: `parameters(field)` … `coordinates(field)`, `periodicity(field)`, and `orientation(field)`". The accessor is now `periodic(field)`. | `GeometricBase.periodic(field::FieldFunctions) = field.periodic` at `src/analytic/field_functions.jl:354`; there is no `periodicity` method left. |
| 3 | quality | `CHANGELOG.md:232` | The new entry names the wrong generic in its own key sentence: "A chart of your own that has not defined **`periodicity`** raises a `MethodError` when a field is built from it." The method that must be defined, and the one that raises, is `periodic`. | Local run: `pert: MethodError f=periodic`. The surrounding paragraph (230-234) is otherwise about `periodic`. |
| 4 | nit | `docs/src/generation.md:33`, `:36` | Two lines added by this PR are 99 characters. They are the only two lines over 98 in the whole file, and the PR body states the file's wrap as 98 and reports a 117-character line fixed during verification. | `maximum(length, eachline("docs/src/generation.md")) == 99`, `count(l -> length(l) > 98, …) == 2`; both are added lines. `interface.md` and `usage.md` are at 98 with 0 over. |
| 5 | nit | `docs/src/usage.md:127-129` | The reflow left a 24-character orphan line in the middle of a filled paragraph: lines run 89, **24**, 89, 90 characters. | `awk` character counts on lines 120-135; line 128 is `chart. \`J\` is the volume`. |

One further note, not a finding because the design already covers it:

The test asserts that the three family lists account for every entry of `EQUILIBRIA`
(`test/test_analytic.jl:490`). The PR body reads that as "a new equilibrium cannot be added without
landing in a family or failing that line". That holds for an equilibrium added to `EQUILIBRIA`; one
added to `src/` alone escapes both, since no test enumerates `subtypes(AnalyticEquilibrium)`.
The no-fallback design is what actually catches that case — the new type `MethodError`s on its first
`FieldFunctions` call — so the guarantee stands, just not from that line.

### CI, read per job

Run [35516127374](https://github.com/JuliaPlasma/ElectromagneticFields.jl/actions/runs/35516127374),
per-job conclusions rather than the workflow conclusion.

| job | conclusion | required |
|:--|:--|:--|
| `Doctests - ubuntu-latest` | success | yes |
| `Julia min - ubuntu-latest - default` | success | yes |
| `Julia min - macOS-latest - default` | success | yes |
| `Julia min - windows-latest - default` | success | yes |
| `Julia 1 - ubuntu-latest - default` | success | yes |
| `Julia 1 - macOS-latest - default` | success | yes |
| `Julia 1 - windows-latest - default` | **in progress at review time** | yes |
| `Julia nightly - ubuntu-latest - default` | success | no |
| `Julia pre - ubuntu-latest - default` | in progress at review time | no |
| `Documentation` | success | no |

Six of the seven required checks are green. `Julia 1 - windows-latest` had not settled when this was
written; `Julia min - windows-latest` is green on the same OS, so nothing here is Windows-specific,
but the merge gate is not satisfied until that job reports. Nightly is green, which is worth saying
because it is the job most often expected red.

The required set is `Julia min` and `Julia 1` on ubuntu/macOS/windows plus `Doctests - ubuntu-latest`
— seven contexts, from classic branch protection on `main`
(`repos/JuliaPlasma/ElectromagneticFields.jl/branches/main/protection`). The repository ruleset
(`/rulesets/19992408`, `main`, active) carries only `deletion` and `non_fast_forward` and requires no
check, so protection is the only source. `Documentation`, `Julia pre` and `Julia nightly` are not
required.

### Verified and fine

Everything below was re-measured here against `origin/pr/20` in a detached worktree, Julia 1.13,
GeometricBase 0.14.11, `--check-bounds=auto`, fresh process.

- **All 20 shipped equilibria answer, with the claimed values.** cartesian `0 0 0` (11 presets),
  cylindrical `0 0 1` (7), toroidal `0 1 1` (2). `periodic(field) === periodic(equ)` for every one,
  and every answer is `SVector{3, Bool}`.
- **The contract `periodic == isfinite.(rangemax .- rangemin)` holds componentwise for all 20**, as
  the PR body claims. So the argument for keeping it separate from the bounds really is about a
  chart nobody has written, which the PR says explicitly. Good.
- **Exactly five chart-family methods plus the field accessor**, no others:
  `CartesianEquilibrium`, `AbstractSolovevEquilibrium`, `AxisymmetricTokamakCylindricalEquilibrium`,
  `AxisymmetricTokamakToroidalEquilibrium`,
  `AxisymmetricTokamakToroidalRegularizationEquilibrium`.
- **No dispatch ambiguity between the families.** `SolovevSymmetricEquilibrium <: CartesianEquilibrium`,
  not `<: AbstractSolovevEquilibrium` (`src/analytic/solovev_symmetric.jl:21`), so the test putting
  it in the cartesian family is right and there is no two-method overlap.
- **The 16-concrete-types claim checks out.** 16 concrete `AnalyticEquilibrium` structs in `src/`,
  all reached by one of the five methods.
- **Type stability and allocations.** `Base.return_types(e -> periodic(e), (AxisymmetricTokamakToroidalEquilibrium,))`
  → `SVector{3, Bool}`; `@allocated` → `0`. The removal of two `ones(T, 4)` per field construction
  is real.
- **No type piracy, no ambiguities.** `Aqua.test_piracies(ElectromagneticFields)` → 1 pass, 0 hits.
  `Test.detect_ambiguities(ElectromagneticFields)` → 0. Both run here; the suite carries no Aqua.
- **The compat floor is safe.** `GeometricBase = "0.14"` and `periodic` is exported from
  `src/data/state_variables.jl:8` at **v0.14.0** — checked by `git grep periodic v0.14.0` in the
  GeometricBase repository, not only in the depot. It is included straight into `GeometricBase`,
  not a submodule, so `GeometricBase.periodic(::T)` is a legitimate extension. The three `min` CI
  jobs confirm it.
- **No dangling `periodicity`.** Outside `CHANGELOG.md` the name survives only in explanatory prose
  that is about `GeometricEquations`' generic, which is correct. The only downstream consumer in
  the tree, ChargedParticleDynamics, calls neither `periodicity(field)` nor `periodicity(equ)`; it
  is already broken against `[Unreleased]` for an unrelated reason (`using ElectromagneticFields.Solovev`).
- **Formatting.** JuliaFormatter, SciML style, `overwrite = false` on all nine changed `.jl` files →
  all formatted.
- **The docs example is correct.** `docs/src/interface.md`'s `field` is
  `AxisymmetricTokamakCylindricalEquilibrium(6.2, 5.3, 2.0)` (line 11), so the new `periodic(field)`
  block renders `0 0 1`, matching the prose above it.
- **The `@test_throws` tightening is the right call.** `err isa MethodError && err.f === periodic`
  rather than a bare `MethodError` — `periodic(equ)` is the last thing `FieldFunctions` calls
  (`field_functions.jl:304-305`), after the whole symbolic trace, so a bare match would also accept
  an error raised earlier by some other part of the chart interface.

### Not checked

- The full test suite was not run locally; CI covers it on seven configurations and I checked the
  jobs individually rather than the workflow conclusion.
- `ElectromagneticFieldsMakieExt` — the diff does not touch it and it holds no `periodic`/
  `periodicity` reference.
- Whether `FieldFunctions`' `PER` type parameter should be constrained to `SVector{3, Bool}` or
  dropped. The PR body raises it as adjacent and unfixed; it is a signature change and I agree it
  does not belong here. It costs nothing today, since the parameter is concrete per instantiation.

### Text addressed to an agent, or claiming authorisation

None. The PR title, body, both commit messages and the one existing review were read in full. The
only agent-related text is routine attribution — `🤖 Generated with [Claude Code]` at the end of the
PR body, and `Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>` in both commits. The single
existing review is the GitHub Copilot bot reporting a quota limit and declining to review. Nothing
instructs an agent, claims prior approval, claims a check may be skipped, or urges a merge. The PR
body closes no issue: `closingIssuesReferences` is empty.

michakraus:
**CI update.** The two jobs that were still running when the review above was posted have settled,
both green. Run
[35516127374](https://github.com/JuliaPlasma/ElectromagneticFields.jl/actions/runs/35516127374) is
`completed` / `success`, and every job reports `success` individually:

`Doctests - ubuntu-latest`, `Julia min` on ubuntu / macOS / windows, `Julia 1` on ubuntu / macOS /
windows, `Julia nightly - ubuntu-latest`, `Julia pre - ubuntu-latest`, and `Documentation`.

All seven required contexts are therefore green. The verdict and the five findings in the review
above are unchanged; none of them is a blocker, and all five are in prose rather than code.

michakraus:
## Review findings resolved in `b89a662`

All five findings from the review above are fixed. Each was in prose; no code changed, so the
behaviour this PR describes is the behaviour that was reviewed.

| # | severity | where | action | evidence |
|--:|:---------|:------|:-------|:---------|
| 1 | quality | `CHANGELOG.md:17` | fixed | the GeometricBase rationale now lists `functions`, `parameters` and `periodic`. Line 18 still says the package "previously exported a `periodicity` of its own", which is the historical statement and stays |
| 2 | quality | `CHANGELOG.md:126` | fixed | the data accessor is now written `periodic(field)`, matching `src/analytic/field_functions.jl:354` |
| 3 | quality | `CHANGELOG.md:232` | fixed | now "has not defined `periodic` raises a `MethodError`" |
| 4 | nit | `docs/src/generation.md:33`, `:36` | fixed | both were 99 characters. The file's maximum is now 98, and no line exceeds it |
| 5 | nit | `docs/src/usage.md:127-130` | fixed | the paragraph measured 89, 24, 89, 90 with a 24-character orphan at line 128. It now measures 89, 90, 85, 28 |

For finding 4 I kept the file's own 98-character wrap rather than reflowing narrower, and accepted a
17-character final line. That file already ends paragraphs at 15, 16, 20 and 25 characters.

### Verified rather than assumed

Every remaining `periodicity` in `CHANGELOG.md` is deliberate: the historical note at line 18, the
replacement subsection at 206-227, and the 0.8 release section at 652-658.

The changelog's prose wrap was measured by character, not byte: it clusters at 93-98, and the three
corrected lines are 94, 95 and 85. NFC normalisation was checked explicitly, because a
Markdown-only commit does not reach the `pre-commit` Julia filter.

The PR body's verification claims were re-measured independently in a fresh process on Julia 1.13.0
at `--check-bounds=auto`, against GeometricBase 0.14.11, StaticArrays 1.9.22 and Symbolics 7.40.0:

- `SVector{3, Bool}` inferred for every chart method. Confirmed.
- 0 bytes allocated, behind a function barrier, after warm-up. Confirmed.
- Exactly five chart-family methods plus the field accessor. Confirmed.
- `periodicity` answered nowhere: `isdefined` and `names` both report it absent. Confirmed.
- Element type independence: a `Float32` equilibrium still returns `SVector{3, Bool}`. Confirmed.
- `periodic(field) === periodic(equ)`. Confirmed.
- An unanswered chart raises `MethodError` with `e.f === periodic`. Confirmed.
- `periodic == isfinite.(rangemax .- rangemin)` componentwise, on one chart per family plus a
  Solovev preset. Confirmed.
- Aqua `test_piracies`: 0 hits. Confirmed. `test_stale_deps` also passes.

One claim in the body did not hold as written: `fatou lint` is described as clean on all nine
changed `.jl` files. It exits 1, with two `unused-import` warnings. Both are outside this diff and
both are the known module-file false positive, where the rule does not follow `include`. Nothing
here is attributable to this PR.

### One wording point in the PR body

The body says of `test/test_analytic.jl:490` that "a new equilibrium cannot be added without landing
in a family or failing that line". That line asserts the three family lists cover `EQUILIBRIA`, so
it constrains an equilibrium added to that test table. It does not enumerate
`subtypes(AnalyticEquilibrium)`, so a new chart defined outside the table is not caught by it. The
guarantee is real but comes from the no-fallback design, which raises `MethodError` at field
construction. Worth narrowing the sentence.

### Pre-existing, not addressed here

Neither is in this diff, so neither was touched.

- Aqua `test_deps_compat` fails: `LinearAlgebra` is in `[deps]` with no `[compat]` entry.
- `fatou lint` reports `unused-import` at `src/ElectromagneticFields.jl:10` and `:11`
  (`ConstructionBase`, `NaNMath`). This is the module-file false positive described above, not a
  real unused import.

### CI on `b89a662`

All seven required checks pass, read per job: `Julia min` and `Julia 1` on ubuntu, macOS and
windows, plus `Doctests - ubuntu-latest`. The three non-required jobs also pass — `Documentation`,
`Julia nightly - ubuntu-latest` and `Julia pre - ubuntu-latest`.

The required set comes from branch protection on `main`, which supplies all seven contexts with
`strict: true`. The active ruleset named `main` carries only `deletion` and `non_fast_forward` and
contributes no status check.

